Issue #46, opened 2016-02-10
Create Test Environment Compute Server


sheyd:
Mirror Data 1.0 data.compute server.

Steps to Complete:
- [x] Bring up server
- [x] Create Ansible playbook from packages in Prod Data 1.0 Compute Server
- [x] Create User with credentials for Quantum
- [x] Test Setup
- [x] ?
- [x] Data!

sheyd:
Need env variable setup to test DB connections. Blocking #67

sheyd:
@namimody Quick update. I have to put in a test branch to re-factor some code paths for the old ETL script repo that Josh put together. It shouldn't take too long. Just wanted to let you know it's another step I'm working on right now.

sheyd:
@namimody TESTED WORKING!!! AHHHHHH SO EXCITED!!!

PR #60, opened 2016-02-19
Add NetCat to support bastion host proxy.

sheyd:
#### What's this PR do?
* Add `netcat` package to bastion host playbook to allow SSH proxy.

* bastion-host.yml file.

* Already tested for `qm1`

* Moving towards not storing private keys on servers, as well as allowing users to directly SSH to server they need to.

#46

### DIFF
--- a/ansible/bastion-host.yml
+++ b/ansible/bastion-host.yml
@@ -3,3 +3,7 @@
   become: true
   roles:
     - base-os
+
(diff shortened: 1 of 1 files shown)


sheyd:
@DFurnes Tested this working. So... auto-merging. Just an FYI ;)

Issue #67, opened 2016-02-26
ETL Process for Mobile_Campaign_Lookup Table


sheyd:

sheyd:
Dug into this yesterday, and it's a tad more complicated than it appears on the surface. @namimody I've upgraded this to "medium". I got lost in coding rabbit-hole yesterday. I'm going to hack away at this some more today. I should have a better update by EOD about if I'm going to be able to complete this today/tomorrow, or if it's going to take longer/I should figure out a different tactic.

sheyd:
Blocked by #46

sheyd:
And... FIRST ROUND FINISHED!!!

I've confirmed that the opt-in-path count from the API matches the number of records in the `mobile_campaign_id_lookup` table.

Current bug is certain campaigns have one number in the "description" field that doesn't actually represent the NID, like `opt_in_path 116871`. 

@Bladt wants to have a further spot check done on this to ensure data fidelity. @namimody I'll leave you to coordinate this with Jeff. I think this ticket can be closed. I'm committing the code now.

PR #79, opened 2016-03-18
Adding global etl-scripts run path.

sheyd:
#### What's this PR do?
Takes `etl-scripts` path to be absolute from server root, instead of user home directory.

Below.

Already done.

`etl-scripts` have some relative and home directory path assumptions that are no longer valid.

#67.

#### Questions:
Am I just going to approve my own PR?


sheyd:
Why yes, yes, I am.

Issue #107, opened 2016-04-21
Create and Document Process for Setting Up Folks in Hubble via SQL Pro


namimody:

sheyd:
Setup instructions in LastPass for internal users and shared with Data Team. @namimody Did you keep a list from our meeting of what users I need to add?

namimody:
@sheyd yes -- do you need names or email addresses

sheyd:
Names, please!

namimody:
James
Hannah
Fantini
Chris
Hilary

PR #109, opened 2016-04-21
Updating bastion hubble host to incorporate data team members.

sheyd:
#### What's this PR do?
* Updates Hubble bastion host to use latest security practices and http/2 protocol.
* Corrects syntax for variable evaluation of `open_access`.
* Adds `open_access` users to Hubble bastion host.

Below!
Already tested and deployed to Hubble Prod.
#107


sheyd:
@DFurnes ?

ansible/hubble-bastion.yml
```diff
@@ -3,7 +3,9 @@
   become: true
   roles:
     - base-os
-    - nginx
+    - nginx-secure
+  vars:
+    open_access: true
```
DFurnes: Not sure if this is a standard Ansible thing, but if not would it make sense to add a comment explaining what this does with regards to creating users and SSH keys below?

sheyd: Notes are usually left in Ansible role, but comments definitely can be added to playbooks. I'll clean-up and add.

DFurnes:
👍

ansible/hubble-bastion.yml
DFurnes: Cool cool. 👍 ✏️

Issue #259, opened 2016-11-04
Move all ETL Scripts to Quasar Repo 


jessleenyc:
- [ ] Update DB Connector to Postgres
- [ ] Updates queries to Postgres

@sheyd please list out the scripts we're keeping

sheyd:
All done!

PR #266, opened 2016-11-15
Adding refactored MailChimp Subscribers Script for RedShift.

sheyd:
#### What's this PR do?

* Moves all ETL Scripts currently run in Hubble to Quasar Public repo.
* All scripts updated to use Python 3.
* Keeping the config in the same path as repo, but private, so added to .gitignore.
* All scripts that were written by @joshcusano have been pre-pended with `jc_`.
* Re-factored `run_update.sh` script to use Python MySQL wrappers.

All the files below. @mshmsh5000 I've sent you original ETL repo link via Slack so you can compare between the two.

* Tested in Quasar.

https://github.com/DoSomething/quasar/issues/259


sheyd:
@mshmsh5000 Given that Furnes is out, would you mind be my second set of 👀  on this and the forthcoming ETL PR's?

sheyd:
@mshmsh5000 actually, ignore this, going to update the merge logic a bit so duplicate records don't make it into the table, via this Redshift guide (so primary table doesn't have to be destroyed over and over): http://docs.aws.amazon.com/redshift/latest/dg/merge-replacing-existing-rows.html. Stand-by for update.

mshmsh5000:
Hokey dokes

sheyd:
@mshmsh5000 Can you take a look now?

sheyd:
@mshmsh5000 Give me one more minute, there might be dupes possible with my re-factored logic. I think I need a distinct statement in there.

sheyd:
@mshmsh5000 Okay, finally got it figured out. Please check the two scripts now :). Tested them in Quasar Hubble tables, and things look good.

sheyd:
@mshmsh5000 Re-factored scripts in place. Please review are you earliest convenience.

etl-scripts/create_mobile_master_lookup_lite.py
```diff
@@ -0,0 +1,49 @@
+import sys
+import config
+import MySQLdb
+import MySQLdb.converters
+
+### Create DB Connection with Appropriate Conversions
+#   Conversions inherited from Josh's other ETL scripts.
+conv_dict = MySQLdb.converters.conversions.copy()
```
mshmsh5000: Asking out of ignorance: Do we still need all of this `converters` business?

etl-scripts/jc_mailchimp_subscribers.py
```diff
@@ -0,0 +1,83 @@
+import json
```
mshmsh5000: This script seems to have nontrivial logic differences from `source_pulls/mailchimp/mc_subscribed.py`, which I assume is its precursor. I assume this is all of the changes to page through MailChimp API results. Anything else?

etl-scripts/jc_mailchimp_subscribers.py
```diff
@@ -0,0 +1,83 @@
...
+origin_time = time_now - backfill_time
+
+# MailChimp API v3.0 HTTP Simple Auth Credentials
+un = config.mailchimp_api_user
+pw = config.mailchimp_api_pass
+
+# Set Initial Offset of O
+member_offset = 0
+
+# MailChimp API v3.0 Paremeters to Get Subscribed Users over designated time window by paginating via "offset" parameter.
```
mshmsh5000: Nit-pick: sp: "parameters"

etl-scripts/jc_mailchimp_unsubscribers.py
```diff
@@ -0,0 +1,79 @@
...
+origin_time = time_now - backfill_time
+
+# MailChimp API v3.0 HTTP Simple Auth Credentials
+un = config.mailchimp_api_user
+pw = config.mailchimp_api_pass
+
+# Set Initial Offset of O
+member_offset = 0
+
+# MailChimp API v3.0 Paremeters to Get Unsubscribed Users over given time period and iterating over 1000 users and paginating
```
mshmsh5000: Sp: parameter (lines 35-36).

etl-scripts/jc_run_update.py
```diff
@@ -0,0 +1,205 @@
+import MySQLdb
```
mshmsh5000: This appears to be a rewrite of the old `etl-scripts/new_etl_process/runupdate.sh`, in which case, 👍

etl-scripts/create_mobile_master_lookup_lite.py
sheyd: @mshmsh5000 I don't strictly know, but the next phase of our ETL scripting (that I've discussed a bit with @DFurnes) is abstracting out two key things:

* DB connectivity
* HTTP scraping with credential passing.

Going forward, I'll also have a lot better grasp of data types we use for new tables, so want to not touch these if they might break tables we already have. But expect them to be deprecated going forward.

etl-scripts/jc_mailchimp_subscribers.py
sheyd: Yes, that's it. The first version we tested and used is in https://github.com/DoSomething/quasar/blob/etl-hubble-v1/etl-scripts/mailchimp_subscribers_daily_snapshot.py, which I wrote the logic for essentially from the ground up. And we've tested it multiple times with confirmed results.

sheyd:
@mshmsh5000 Updated `paremeter` issues and other comments. Quick 👀 before I merge in?

mshmsh5000:
👍 🌮

PR #269, opened 2016-11-29
Modified profile API test script to not use Python 2 "long" type.

sheyd:
#### What's this PR do?

* Updated Profile Update ETL script to not use "long" type in two if loops, since the "int" type in Python3 covers all integer types.

The one file here.

Limited testing done in Quasar. (*For the love of all that is holy I need a local dev environment*).


sheyd:
@mshmsh5000 Merging this in since it's a blocker, but wanted to my Python Peer in case we need to reference this at any point.

Issue #273, opened 2016-11-29
Prototype First Version of User Activity Table


sheyd:
_TASK_
-------------------
#### Ideal Deadline

12/1/2016

#### Request

Create activity log by diffing daily `dosomething.users` table copy and diffing previous days and create user changes.

#### Why This Matters/Use Case Example

We want to track up to date user behaviour, and this will help the Data team track overall changes of users through the system as an MVP.

jessleenyc:
Creates unnecessary joins: 
First Name
Phone Number
Source
Address

sheyd:
@nmccormick619 and @Bladt now have access to the Quasar DB user log POC table to play around with.

jessleenyc:
@sheyd Can we close this?

PR #278, opened 2016-12-14
Adding first version of Phoenix User Log POC creation

sheyd:
#### What's this PR do?

Adds first version of POC User Log sources from Phoenix DB.

The one script!

Already tested in Prod.

https://github.com/DoSomething/quasar/issues/273

### DIFF
--- a/etl-scripts/update_phoenix_user_log.py
+++ b/etl-scripts/update_phoenix_user_log.py
@@ -11,14 +11,14 @@ db = MySQLdb.connect(host=config.host, #hostname
 
 cur = db.cursor()
 
-# Record start time
+# Record start time.
 start_time = time.time()
 
-# Get List of All Phoenix Staging Tables
+# Get list of all Phoenix DB staging tables.
 cur.execute("SHOW TABLES LIKE '%_staging_changes';")
 db.commit()
 
-# Delete all pre-existing staging tables for re-runs of script
+# Delete all pre-existing staging tables for re-runs of script.
 old_staging_tables = cur.fetchall()
 for x in old_staging_tables:
     staging_table_name = str(x).strip('(').strip(')').strip(',').strip("'")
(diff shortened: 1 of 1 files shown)


sheyd:
@mshmsh5000 or @DFurnes Would one you mind 👀 this before pushing live?

etl-scripts/update_phoenix_user_log.py
```diff
@@ -0,0 +1,77 @@
...
+    db.commit()
+
+# Get List of All Phoenix Snapshots
+cur.execute("SHOW TABLES;")
+db.commit()
+
+# Iterate over tables and print name, testing
+table_snapshot = cur.fetchall()
+for x in table_snapshot:
+    table_name = str(x).strip('(').strip(')').strip(',').strip("'")
```
mshmsh5000: What's the original table name format that you're cleaning up?

sheyd:
@mshmsh5000 Can you 👀  this PR so I can automate our User Table POC?

mshmsh5000:
@sheyd Sorry, I had [that one question](https://github.com/DoSomething/quasar/pull/278#discussion_r92691105) for you back when you posted the PR, and never saw a response.

etl-scripts/update_phoenix_user_log.py
```diff
@@ -0,0 +1,77 @@
...
+          passwd=config.pw, #  password
+          db='phoenix_user_snapshots' # source and staging area for snapshots
+          )
+
+cur = db.cursor()
+
+# Record start time
+start_time = time.time()
+
+# Get List of All Phoenix Staging Tables
```
mshmsh5000: Super nit-picky, but these comments are usually just sentences—don't need to be all capitalized words.

mshmsh5000:
The rest of this makes sense to me, just the general note on comment formatting for everyone's benefit:

```
# This Comment Doesn't Need to be All Upper Case
```

should generally be

```
# This comment doesn't need to be all upper case
```

or, if you're me,

```
# This comment doesn't need to be all upper case.
```

sheyd:
@mshmsh5000 Updated syntax changes and answered your questions inline with comment.

mshmsh5000:
This is python trivia, but you should be able to strip out any chars with a single command, as long as they're all at the beginning and end of the string. I think you'd get the same result from the shorter

```
table_name = str(x).strip("(),;'")
```

mshmsh5000:
Anyhoo, 👍

Issue #280, opened 2017-01-03
Create Quasar Web Scraping Python Library


sheyd:
_TASK_
-------------------
#### Ideal Deadline
1/9/17

#### Request
Most of our ETL scripts have bad code re-use mechanisms involving copying and pasting, especially for web scraping. 

The goal is to provide a unified web-scraping library with built-in methods for:
1. Specifying retries (with sane retry defaults assumed).
2. Handling HTTP Simple Auth
3. Handling OAuth2 Auth (first use case auth'ing to Northstar)
4. Specifying endpoints and URL params easily.
5. Some Sane De-bugging, status returns.

#### Additional Information
All of the above functionality might not make it into the first version of the library, but at minimum, the first 3 points must be addressed.

#### Why This Matters/Use Case Example
One of the premises for the Quasar Platform is preferring web/API scraping to direct DB access for data ingestion. We need a well-defined and engineered library to make this easy.

Tagging @DFurnes for sanity and guidance on implementing this correctly.

sheyd:
Talked to @DFurnes and @jessleenyc I was wrong about potentially not using Gateway as a template. There's a few different ways to consider architecting the web-scraper library, and Furnes encouraged me to look at Gateway as a template or guide of sorts. Researching it now.

sheyd:
There are two options for use with OAuth 2 for Python. `oauthlib2` plus `requests` vs Google's `oauth2client`. I'm spinning up a VM to test both with requests to Northstar endpoints to see if it works.

sheyd:
Got things working with `oauthlib2`! Testing oauth2client shortly.

sheyd:
Going to go with `oauthlib2`. `oauth2client` is pretty powerful, but a bit more Google focused than I like. We can always swap out libraries once the essentials are defined anyways.

PR #287, opened 2017-02-06
Northstar and Quasar Web Scraper

sheyd:
#### What's this PR do?

* Defines base Web Scraper class that takes retries into account in case of network failure.
* First version of DS Northstar Scraper (first DS API specific Python Quasar library!)

* [Here](https://github.com/DoSomething/quasar/compare/the-e-in-etl?expand=1#diff-b2a7d92d0454a839532f4d59746f3707)
* [Then Here]( https://github.com/DoSomething/quasar/compare/the-e-in-etl?expand=1#diff-9f9baf68a7466e623c7cd851bdba229a)

* If proper Python 3 libraries and virtaulenv are setup, then creating test script with proper config file to get some JSON from Northstar!

* This is the scraper script that will be used for the ETL job of populating new Quasar User table schema.

* #280

### DIFF
--- a/etl-scripts/DSNorthstarScraper.py
+++ b/etl-scripts/DSNorthstarScraper.py
@@ -6,6 +6,7 @@ from requests_oauthlib import OAuth2Session
 from QuasarWebScraper import Scraper
 
 class NorthstarScraper(Scraper):
+    """Class for extracting Northstar data via API."""
 
     # Define OAuth2 Client ID and Client Secret.
     ns_client = config.ns_client_id
@@ -16,27 +17,34 @@ class NorthstarScraper(Scraper):
     oauth = OAuth2Session(client=client)
 
     def __init__(self, ns_url='https://profile.dosomething.org'):
+        """Set Northstar API with all the retry goodness of Quasar Web Scraper."""
         Scraper.__init__(self, ns_url)
 
-    # Send OAuth request for new or refresh token.
+
     # Note to self, should update this method to check for validity of token, and only refresh when necessary.
     # Something like if simple get().status = 422, then new_token, else, return existing token.
     def getToken(self, path='/v2/auth/token'):
+        """Send OAuth request for new or refresh token."""
         new_token = self.oauth.fetch_token(self.url + path, client_id=self.ns_client, client_secret=self.ns_secret, scope='admin')
         return new_token['access_token']
 
-    # Override "get" method from QuasarWebScraper to include OAuth 2 token for authorization.
     def get(self, path, query_params=''):
+        """Set get method to include OAuth 2 Token for Authorization."""
         auth_headers = {'Authorization': 'Bearer ' + str(self.getToken())}
         response = self.session.get(self.url + path, headers=auth_headers, params=query_params)
         return response.json()
 
-    # Get all users with max Nortstar user page. Default to first page of results.
     def getUsers(self, users=100, page_number=1):
+        """Get users in max batch size by default from Northstar API.
+
+        Args:
+            users (int): Users per page returned, default 100, max 100.
+            page_number (int): Page number to return, default 1.
+        """
         user_response = self.get('/v1/users', {'limit': users, 'page': page_number})
         return(user_response['data'])
 
-    # Get total user count and pages. Returns array with member_count and total_pages.
     def userCount(self, users=100, page_number=1):
+        """Get total user count and pages."""
         get_count = self.get('/v1/users', {'limit': users, 'page': page_number})
         return(get_count['meta']['pagination']['total'], get_count['meta']['pagination']['total_pages'])
--- a/etl-scripts/DSNorthstarScraper.py
+++ b/etl-scripts/DSNorthstarScraper.py
@@ -10,7 +10,7 @@ from QuasarWebScraper import Scraper
 class NorthstarScraper(Scraper):
 
     def __init__(self):
-        url = 'https://profile.dosomething.org'
+        url = 'https://profile.dosomething.org/'
         Scraper.__init__(self, url)
 
     # Define OAuth2 Client ID and Client Secret.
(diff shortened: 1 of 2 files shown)


mshmsh5000:
I think it's worth commenting these out using [standard Pythonic commenting practice](https://google.github.io/styleguide/pyguide.html?showone=Comments#Comments). Unfortunate that Python docstring style is ugly, but it's a standard.

Unless there's another standard you think we should adhere to, in which case please find a reference we can review. But let's pin this down (& lint against it!) now that we're making more general-purpose code.

etl-scripts/DSNorthstarScraper.py
```diff
@@ -0,0 +1,57 @@
...
+    def __init__(self):
+        url = 'https://profile.dosomething.org/'
+        Scraper.__init__(self, url)
+
+    # Define OAuth2 Client ID and Client Secret.
+    ns_client = config.ns_client_id
+    ns_secret = config.ns_client_secret
+
+    # Create Client Object and Obtain Token
+    client = BackendApplicationClient(client_id=ns_client)
```
DFurnes: I'm not sure what standard Python style is, but it might be more readable to put class attributes like this above the `__init__` constructor. Otherwise it looks like an indentation mistake at first glance.

etl-scripts/DSNorthstarScraper.py
```diff
@@ -0,0 +1,57 @@
...
+    oauth = OAuth2Session(client=client)
+    # Stub initial token request to be used in getToken method improvement below.
+    # token = oauth.fetch_token(token_url='https://profile.dosomething.org/v2/auth/token', client_id=ns_client, client_secret=ns_secret, scope='admin')
```
DFurnes: It looks like this can be removed. 👾

etl-scripts/DSNorthstarScraper.py
```diff
@@ -0,0 +1,57 @@
+import json
+import config
+import oauthlib
+from oauthlib.oauth2 import BackendApplicationClient
+from requests_oauthlib import OAuth2Session
+from QuasarWebScraper import Scraper
+
+
+# Setup new scraper class to Use Mailchimp endpoint and basic auth.
```
DFurnes: 📝 🙈

etl-scripts/DSNorthstarScraper.py
```diff
@@ -0,0 +1,57 @@
...
+        user_response = self.get('/v1/users', {'limit': users, 'page': page_number})
+        return(user_response['data'])
+
+    # Get total user count. Using 1 page and 1 user to minimize request response time.
+    def userCount(self, users=1, page_number=1):
+        get_count = self.get('/v1/users', {'limit': users, 'page': page_number})
+        return(get_count['meta']['pagination']['total'])
+
+    # Get total pages based on getUsers default values (can be overridden).
+    def pageCount(self, users=100, page_number=1):
```
DFurnes: This is probably something that makes sense to track when making the actual request, since the number of pages could change between requesting this `pageCount()` and then iterating over the actual results from `getUsers()`.

Not necessary for merge but could be an unexpected 🐛 in the future!

etl-scripts/DSNorthstarScraper.py
```diff
@@ -0,0 +1,57 @@
...
+from oauthlib.oauth2 import BackendApplicationClient
+from requests_oauthlib import OAuth2Session
+from QuasarWebScraper import Scraper
+
+
+# Setup new scraper class to Use Mailchimp endpoint and basic auth.
+class NorthstarScraper(Scraper):
+
+    def __init__(self):
+        url = 'https://profile.dosomething.org/'
```
mshmsh5000: Worth providing setters for these URL values, so you can override (for testing, etc.)? Or is that not in the cards?

etl-scripts/DSNorthstarScraper.py
```diff
@@ -0,0 +1,57 @@
...
+    client = BackendApplicationClient(client_id=ns_client)
+    oauth = OAuth2Session(client=client)
+    # Stub initial token request to be used in getToken method improvement below.
+    # token = oauth.fetch_token(token_url='https://profile.dosomething.org/v2/auth/token', client_id=ns_client, client_secret=ns_secret, scope='admin')
+
+    # Send OAuth request for new or refresh token.
+    # Not to self, should update this method to check for validity of initial token above, and only refresh when necessary.
+    # Something like if simple get().status = 422, then new_token, else, return existing token.
+    def getToken(self):
+        new_token = self.oauth.fetch_token(token_url='https://profile.dosomething.org/v2/auth/token', client_id=self.ns_client, client_secret=self.ns_secret, scope='admin')
```
mshmsh5000: Same here. Worth being able to override the `token_url` for testing?

etl-scripts/DSNorthstarScraper.py
```diff
@@ -0,0 +1,57 @@
...
+    # Override "get" method from QuasarWebScraper to include OAuth 2 token for authorization.
+    def get(self, path, query_params=''):
+        auth_headers = {'Authorization': 'Bearer ' + str(self.getToken())}
+        response = self.session.get(self.url + path, headers=auth_headers, params=query_params)
+        return response.json()
+
+    # Stub class for parsing any response as JSON. Currently has some parsing issues.
+    def processResponse(self, response):
+        parsed_response = response.json()
+        return parsed_response
```
mshmsh5000: I'd imagine these lines can get reduced to 

```
return response.json()
```

etl-scripts/QuasarWebScraper.py
```diff
@@ -0,0 +1,42 @@
...
+
+# The code below is re-purposed and inspired by
+# http://stackoverflow.com/questions/21371809/cleanly-setting-max-retries-on-python-requests-get-or-post-method
+# and this was object orientified primarily by David Furnes @DFurnes to help me learn
+# more about OOP in general and via Python specifically.
+
+
+class Scraper:
+
+    # Set Default values for retry and backoff times.
```
mshmsh5000: See my PR note on following Python coding style for comments.

etl-scripts/QuasarWebScraper.py
```diff
@@ -0,0 +1,42 @@
...
+
+        # Create adapters with the retry logic for each
+        http = requests.adapters.HTTPAdapter(max_retries=http_retries)
+        https = requests.adapters.HTTPAdapter(max_retries=https_retries)
+
+        # Replace the session's original adapters
+        self.session.mount('http://', http)
+        self.session.mount('https://', https)
+
+    # Define Basic Get/Post Logic
```
mshmsh5000: Don't need all caps here.

etl-scripts/QuasarWebScraper.py
```diff
@@ -0,0 +1,42 @@
...
+class Scraper:
+
+    # Set Default values for retry and backoff times.
+    def __init__(self, url, retry_total=6, backoff_time=300):
+        # Create a session
+        self.session = requests.Session()
+        self.url = url
+
+        # Define retries and backoff time for http and https urls
+        http_retries = Retry(total=retry_total, backoff_factor=backoff_time)
```
mshmsh5000: Not sure I understand the values passed for `backoff_factor` here. The math used to calculate sleep times between retries is defined as

```
{backoff factor} * (2 ^ ({number of total retries} - 1))
```

So it's a multiplier, not an absolute number. And it can never exceed `Retry.BACKOFF_MAX`, which is set to `120`. Or am I [misunderstanding this](http://pydoc.net/Python/requests/2.9.1/requests.packages.urllib3.util.retry/)?

sheyd:
@mshmsh5000 Regarding coding style, looks like there's PEP8, which uses some of the # annotation we're all familiar with noted here: http://pep8.org/#block-comments. There's also a pep8 Python library to check for syntax correctness 🎉 . The PEP257 commenting system, AKA doctype linked above seems an extension of this (or a super/subset, not entirely sure of historical progression here). I'm eager to comment things to one standard. Is my assessment of the above correct, or am I missing something. Also, which would you recommend fully?

mshmsh5000:
PEP8 does restate the docstring requirement:

>Write docstrings for all public modules, functions, classes, and methods. Docstrings are not necessary for non-public methods, but you should have a comment that describes what the method does. This comment should appear after the def line.

Following that makes sense. So, when you're writing a class definition, you'd do

```
class Baz:

    """
    Crumple all the `foo` available. Levels of crumple controlled via the `depth` param in the constructor.
    """

    depth = 5
    """Default levels of crumple"""
```

...and so on.

etl-scripts/QuasarWebScraper.py
sheyd: I didn't actually know about BACKOFF_MAX values! Thanks for the catch. I fat-fingered 300, should have been 30, which is what our other MoCo scripts should be set to, but that's still way outside the bounds here. Changed to use 6 retries, and 1.9 second backoff, which *should* hit just under 120 for max.

etl-scripts/QuasarWebScraper.py
sheyd: Fixed in latest commit.

Issue #292, opened 2017-02-10
ETL Campaign info from Phoenix to Quasar by 02/17


jessleenyc:
_TASK_
-------------------
#### Request

- [x] Do necessary setup work for pulling campaign data into Quasar. 
- [x] Set up db connection
- [x] Write scripts for ingestion
- [x] ETL 

#### Why This Matters/Use Case Example
Pulling in campaigns info is an MVP for utilizing quasar in dashboards.

sheyd:
Data imported and in production. Punting on `campaign_url` field until we're in Phoenix Next. Confirmed with Jeff this is acceptable, and can actually be recreated in Looker.

PR #312, opened 2017-03-15
Campaign etl from phoenix

sheyd:
#### What's this PR do?
* Adds Phoenix Scraper code we worked on so it's not lost to the vistages time and if it's useful to anyone.
* Runs complete `campaign_info` ETL query against Phoenix into Quasar.

Files here.

Scraper can be manually tested with proper credentials. DB query tested working.

#292


sheyd:
@mshmsh5000 @DFurnes Can one of you review this today? I'd like to get it pushed up into Prod before too long. One stylistic thing is how I wrapped the gnarly SQL query. It's indented a bit improperly, but I was torn between doing that for 79 character PEP8 guidelines or making it a bit more readable. I defaulted to PEP8, but it does look a bit goofy.

Thoughts?

etl-scripts/phoenix_to_campaign_info_table.py
```diff
@@ -0,0 +1,94 @@
+import MySQLdb
+import config
+import time
+import sys
+
+
+start_time = time.time()
+"""Keep track of start time of script."""
+
+db = MySQLdb.connect(host=config.host,  # hostname
```
DFurnes: Not necessary for merging this pull request, but I think it'd make sense to have a `DatabaseClient` class that abstracts this away so you don't have to have it duplicated in each ETL script. 🥞

etl-scripts/phoenix_to_campaign_info_table.py
```diff
@@ -0,0 +1,94 @@
...
+  dosomething.taxonomy_term_data ttd1
+    ON fdfat.field_action_type_tid = ttd1.tid
+LEFT JOIN
+  dosomething.field_data_field_cause fdfc
+    ON fdfc.entity_id = c.field_campaigns_target_id
+LEFT JOIN
+  dosomething.taxonomy_term_data ttd2
+    ON fdfc.field_cause_tid = ttd2.tid
+WHERE c.bundle = 'campaign_run'
+GROUP BY c.entity_id;"""
```
DFurnes: That's one hell of a query.

etl-scripts/phoenix_to_campaign_info_table.py
sheyd: @DFurnes Yeah, I think I'll add that to next quarter's tasks. 💯

Issue #310, opened 2017-03-13
QA Member Count


jamjensen:
_TASK_
-------------------

#### Request
- Once we run the query (#296), we need to confirm that the member count is correct -- if not, this card should be used to document the investigation as to why it's incorrect. 

#### Additional Information
- JB: "Could take anywhere from a single day's worth of work to an entire sprint." 

#### Why This Matters/Use Case Example
- It's a critical goal to have an accurate member count.

sheyd:
NS ingestion done @Bladt! Let the QA'ing begin.

mikefantini:
QA'ing the member count is a blocker for #296

jamjensen:
I just queried the Quasar `users` table for the number of active MoCo subscribers and returned 3,492,915 which is the same exact number I saw on Wednesday. It's also the same number that's in the `all_moco_users` table.  The MoCo GUI currently displays 3,509,362 active subscribers

How often does the script run to update number of active MoCo subscribers in the quasar `users` table?

jamjensen:
DF: "small number of users due to drupal id conflict... members have same drupal id on more than one accounts, which prevents us from updating them"

jamjensen:
Confirming that the number of active MoCo subscribers in the quasar users table is fluctuating based on live updates to Northstar. The difference between the number of active subscribers in the MoCo UI (3,502,427) and the number of active subscribers in quasar users table (3,495,025) is now 7,402.  

Dave is exploring if the association of multiple drupal_ids to different northstar accounts (and the fact that we're unable to update their accounts) can explain the difference. I've also opened a support ticket with MoCo to see if there's any way a user can end up on their Active Subscriber list without having status = active.

jamjensen:
*Update as of 6/6*
- DF to run through another MoCo backfill to see if we skipped over accounts that we didn't properly handle due to issue of duplicate drupal ids.

Issue #294, opened 2017-02-10
Data Analyst able to confirm full Northstar dataset exists in Quasar Users Table


jessleenyc:
_TASK_
-------------------
#### Description
As a Data Analyst, I want to review user data in the Quasar `Users` table so that we can confirm that all Northstar user data has been ingested into the Quasar `Users` table. 

#### Why This Matters
Reviewing the Northstar dataset ingestion is necessary so that we are confident that all user information exists in the Quasar `Users` table. 

#### Definition of Done
-The following checklist is completed: https://docs.google.com/document/d/1xKsmhPamo1RqwofqHsi0fv5O3F-YoPFhQrVS7yCKIx0/edit#

sheyd:
Quasar ingestion is still on-going. Looks like it should be complete by Monday!

sheyd:
User count matches Northstar count! Let the reviews begin!

Bladt:
Reviewing has started. 

Here is a Look for those playing along at home to see Users by NS Create Date and NS Source: https://data.dosomething.org/looks/446?toggle=det,fil

Bladt:
I tried to group by first name. Got this error. Something might be going on with indexing:
![image](https://cloud.githubusercontent.com/assets/4752626/24169735/f30de7a8-0e54-11e7-903f-dcfe5bdb962f.png)

nmccormick619:
What's the deal with the Drupal uptick on 2/08?
<img width="1143" alt="screen shot 2017-03-21 at 5 22 29 pm" src="https://cloud.githubusercontent.com/assets/17914600/24171471/0063fa68-0e5b-11e7-9ca6-0815bd6a6c0b.png">

sheyd:
I've put in a PR ^^ to update user count. There are some comments in there I want to get into this PR that will help things considerably. Also investigating a slowdown in the Northstar scraping again. I'll update in another ticket.

Bladt:
I need to update mgmt-team on when we will be able to run current member count. When do we think we will have rows 24-38 (pre-reqs for a member count) from the schema?

jamjensen:
Blocked by #321

Bladt:
using the following to test member count:

'
select 
  count(case when qu.mailchimp_subscription_status = 'subscribed' or qu.moco_current_status = 'subscribed' then qu.northstar_id end) as 'total_members',
  count(case when qu.mailchimp_subscription_status = 'subscribed' then qu.northstar_id end) as 'email_subs',
[...]

PR #315, opened 2017-03-22
Northstar ETL Update - Fast Mode and Paginate to Complete

sheyd:
#### What's this PR do?
* Changes ETL extract to use "next page" in processing requests from Northstar instead of static limit when ETL process starts. Essentially jobs will run through to end of time period/request than 
missing potential new pages that come in.
* Adds testing for Thor Northstar endpoint as first step towards a QA environment. 
* Updated DS Northstar Scraper to default to `fast` mode.
* Added nextPageStatus method to DS Northstar Scraper to figure out if there are any pages left or not. 
* Added 3 new fields required by Quasar schema. Still missing `updated_at`. Will add via another PR, since it has to scrape the entire Northstar data set again.

https://github.com/DoSomething/quasar/blob/thor-northstar-ETL-test/etl-scripts/DSNorthstarScraper.py

Already tested against Thor Northstar API endpoint as working. 

https://github.com/DoSomething/quasar/issues/294
#310

### DIFF
--- a/etl-scripts/DSNorthstarScraper.py
+++ b/etl-scripts/DSNorthstarScraper.py
@@ -41,10 +41,23 @@ class NorthstarScraper(Scraper):
             users (int): Users per page returned, default 100, max 100.
             page_number (int): Page number to return, default 1.
         """
-        user_response = self.get('/v1/users', {'limit': users, 'page': page_number})
+        user_response = self.get('/v1/users', {'limit': users, 'page': page_number, 'pagination': 'cursor'})
         return(user_response['data'])
 
     def userCount(self, users=100, page_number=1):
         """Get total user count and pages."""
         get_count = self.get('/v1/users', {'limit': users, 'page': page_number})
         return(get_count['meta']['pagination']['total'], get_count['meta']['pagination']['total_pages'])
+
+    def nextPageStatus(self, users=100, page_number=1):
+        """Get users in max batch size by default from Northstar API.
+
+        Args:
+            users (int): Users per page returned, default 100, max 100.
+            page_number (int): Page number to return, default 1.
+        """
+        user_response = self.get('/v1/users', {'limit': users, 'page': page_number, 'pagination': 'cursor'})
+        if user_response['meta']['cursor']['next'] is None:
+            return False
+        else:
+            return user_response['meta']['cursor']['next']
--- a/etl-scripts/DSNorthstarScraper.py
+++ b/etl-scripts/DSNorthstarScraper.py
@@ -44,13 +44,23 @@ class NorthstarScraper(Scraper):
         user_response = self.get('/v1/users', {'limit': users, 'page': page_number, 'pagination': 'cursor'})
         return(user_response['data'])
 
+    def getUsersSlow(self, users=100, page_number=1):
+        """Get users in max batch size by default from Northstar API.
+
+        Args:
+            users (int): Users per page returned, default 100, max 100.
+            page_number (int): Page number to return, default 1.
+        """
+        user_response = self.get('/v1/users', {'limit': users, 'page': page_number})
+        return(user_response['data'])
+
     def userCount(self, users=100, page_number=1):
         """Get total user count and pages."""
         get_count = self.get('/v1/users', {'limit': users, 'page': page_number})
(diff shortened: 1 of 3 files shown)


sheyd:
I'm not going to worry about the 79-line limit rule for this PR, it'll make the requests and DB queries look goofier than they already are.

etl-scripts/northstar_to_user_table.py
```diff
@@ -47,15 +44,14 @@
...
     base_string = str(base_value)
     strip_special_chars = re.sub(r'[()<>/"\'\\]', '', base_string)
     transform_null = re.sub(r'^$|\bNone\b', 'NULL', strip_special_chars)
     return str(transform_null)
 
-while i <= ns_pages:
+while nextPage is True:
     current_page = ns_fetcher.getUsers(100, i)
     for user in current_page:
         query = "REPLACE INTO quasar.users (northstar_id,\
```
mshmsh5000: Not a blocker, but I would DRY up the repeated query here and [below](https://github.com/DoSomething/quasar/blob/18da3d9be13522b5e34f9220c587ed24e9f4d1ed/etl-scripts/northstar_to_user_table.py#L108). Having all that repeated code makes it harder to keep it consistent if you need to change the query in the future. You could just have a 

```
def replace_user(cur, db, user):
```

and extract the code there.

etl-scripts/northstar_to_user_table.py
```diff
@@ -88,15 +88,63 @@ def to_string(base_value):
...
-    cur.execute("REPLACE INTO quasar_etl_status.northstar_ingestion \
-                (counter_name, counter_value) VALUES(\"last_page_scraped\",\
-                \"{0}\")".format(i))
-    db.commit()
+    nextPage = ns_fetcher.nextPageStatus(100, i)
+    print(nextPage)
+    if nextPage is True:
+        i += 1
+        cur.execute("REPLACE INTO quasar_etl_status.northstar_ingestion \
+                    (counter_name, counter_value) VALUES(\"last_page_scraped\",\
```
mshmsh5000: Stickler can be a jerk sometimes.

etl-scripts/thor_northstar_to_user_table.py
```diff
@@ -0,0 +1,153 @@
+import MySQLdb
```
mshmsh5000: Is this just a copy of `northstar_to_user_table.py` with one value changed? If so, (a) all of the comments apply to this file, too, and (b) we should quick find a better way to do this so you can run tests in Thor without maintaining two nearly-identical files.

etl-scripts/DSNorthstarScraper.py
```diff
@@ -41,10 +41,33 @@ def getUsers(self, users=100, page_number=1):
...
+
+    def nextPageStatus(self, users=100, page_number=1):
+        """Check next page pagination to see if null or not.
+
+        Args:
+            users (int): Users per page returned, default 100, max 100.
+            page_number (int): Page number to return, default 1.
+        """
+        user_response = self.get('/v1/users', {'limit': users, 'page': page_number, 'pagination': 'cursor'})
+        if user_response['meta']['cursor']['next'] is None:
```
DFurnes: This doesn't block merging, but we should be passing in [the existing response](https://github.com/DoSomething/quasar/blob/18da3d9be13522b5e34f9220c587ed24e9f4d1ed/etl-scripts/northstar_to_user_table.py#L55) we already have for this check rather than making another separate HTTP request just to check if there's another page.

DFurnes: It'd also be great to use the included `next` URL for loading the following page, since then no matter how you started paging through a result set you can always just grab that pre-made link to the next page.

sheyd:
@DFurnes @mshmsh5000 Need to do some testing with this branch on ETL nodes, so going to merge this in, then a "fix" branch so other processes can keep running uninterrupted... time for me to think about a QA ETL node...

sheyd:
After some simple fixes and comments, of course :)

sheyd:
@mshmsh5000 @DFurnes Merging in cleaned up version. Re-factoring and updates coming in next branch after I track down Thor Northstar response slowdown.

PR #319, opened 2017-03-24
First round of updates:

sheyd:
1. Put in case logic to swap between environments. URL per prod or
thor gets set correctly. Now just have to make sure I set the DB
tables names correctly per env.
2. Cleaned up a bunch of the DB code to take strings as their proper
values and escaped securely. Also cleaned up query sanitizer to return
None types that DB cursor now properly sets to NULL value, not as
as string with value "NULL".
3. Add SSL connectivity for a bit more security over the wire.

Gem of a DB extraction class available here:
http://softwareengineering.stackexchange.com/questions/200522/how-to-deal-with-database-connections-in-a-python-library-module

#### What's this PR do?
1. Put in case logic to swap between environments. URL per prod or
thor gets set correctly. Now just have to make sure I set the DB
tables names correctly per env.
2. Cleaned up a bunch of the DB code to take strings as their proper
values and escaped securely. Also cleaned up query sanitizer to return
None types that DB cursor now properly sets to NULL value, not as
as string with value "NULL".
3. Add SSL connectivity for a bit more security over the wire.

Gem of a DB extraction class available here:
http://softwareengineering.stackexchange.com/questions/200522/how-to-deal-with-database-connections-in-a-python-library-module

Any of the files attached.

Manually tested against Thor.

### DIFF
--- a/etl-scripts/northstar_to_user_table.py
+++ b/etl-scripts/northstar_to_user_table.py
@@ -168,6 +168,9 @@ while nextPage is True:
                                                 to_string(user['source_detail'])))
             db.commit()
 
+cur.close()
+db.close()
+
 end_time = time.time()  # Record when script stopped running.
 duration = end_time - start_time  # Total duration in seconds.
 print('duration: ', duration)
--- a/etl-scripts/northstar_to_user_table.py
+++ b/etl-scripts/northstar_to_user_table.py
@@ -5,9 +5,9 @@ import re
 import sys
 from DSNorthstarScraper import NorthstarScraper
 
-"""DS Northstar to Quasar User ETL script.
+"""DS Northstar to Quasar User ETL Test script.
 
-This ETL scripts scrapes the DoSomething Northstar User API and ETL's the
+This ETL scripts scrapes the DoSomething Thor Northstar User API and ETL's the
 output to our MySQL Quasar data warehouse.
 
 The script takes an optional argument for what Northstar page result to start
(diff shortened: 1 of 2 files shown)


etl-scripts/northstar_to_user_table.py
```diff
@@ -20,48 +20,68 @@
...
+        northstar_env_url = 'https://northstar.dosomething.org'
+    elif sys.argv[1] == 'thor':
+        northstar_env_url = 'https://northstar-thor.dosomething.org'
+    else:
+        print("Please provide a working Northstar environment.")
+        sys.exit(0)
+    i = int(sys.argv[2])
+else:
+    print("Sorry, please specify proper arguments.")
+"""Determine environment, and page to start from.
```
mshmsh5000: Seems to me that a simpler and less error-prone approach would be always to require both arguments. The usage text will be easier to understand, and the invoker will always have to be explicit about page & environment.

etl-scripts/northstar_to_user_table.py
```diff
@@ -20,48 +20,68 @@
...
+ns_fetcher = NorthstarScraper(northstar_env_url)
+
+
+def to_string(base_value):
+    """Converts to string and replaces None values with empty values."""
+    if base_value is None:
+        return None
+    else:
+        base_string = str(base_value)
+        strip_special_chars = re.sub(r'[()<>/"\'\\]', '', base_string)
```
mshmsh5000: Is the exhaustive list of characters you want to strip out? Curious how you derived this set.

etl-scripts/northstar_to_user_table.py
```diff
@@ -74,13 +94,12 @@ def to_string(base_value):
...
                                             moco_current_status,\
                                             moco_source_detail)\
-                                            VALUES(\"{0}\",\"{1}\",{2},\"{3}\",\
-                                            \"{4}\",\"{5}\",\"{6}\",\"{7}\",\
-                                            \"{8}\",\"{9}\",\"{10}\",\"{11}\",\
-                                            \"{12}\",\"{13}\",\"{14}\",\"{15}\"\
-                                            ,NULL,NULL,\"{16}\",\"{17}\",\
-                                            \"{18}\")".format(
-                                            to_string(user['id']),
+                                            VALUES(%s,%s,%s,%s,\
```
mshmsh5000: This looks much cleaner.

etl-scripts/northstar_to_user_table.py
```diff
@@ -20,48 +20,68 @@
...
 
-
+ca_settings = {'ca': '/home/quasar/rds-combined-ca-bundle.pem'}
 db = MySQLdb.connect(host=config.host,  # hostname
                      user=config.user,  # username
-                     passwd=config.pw)  # password
+                     passwd=config.pw,  # password
+                     use_unicode=True,  # Use unicode for queries.
+                     charset='utf8',    # Use UTF8 character set for queries.
+                     ssl=ca_settings)   # Connect using SSL
```
mshmsh5000: Does this throw a useful error if, e.g., the file is missing/unreadable? Would want to be able to debug this quickly if connections were failing.

etl-scripts/thor_northstar_to_user_table.py
```diff
@@ -20,47 +20,68 @@
...
+    else:
+        base_string = str(base_value)
+        strip_special_chars = re.sub(r'[()<>/"\'\\]', '', base_string)
+        return str(strip_special_chars)
 
 while nextPage is True:
     current_page = ns_fetcher.getUsers(100, i)
     for user in current_page:
-        query = "REPLACE INTO quasar.thor_users (northstar_id,\
+        cur.execute("REPLACE INTO quasar.thor_users (northstar_id,\
```
mshmsh5000: So table names like this still need to get altered per environment, right?

sheyd: @mshmsh5000 Yeah. I think I can do a simple `%s` substitution per environment, but haven't tested it yet. Probably will this morning and add to this PR.

etl-scripts/northstar_to_user_table.py
sheyd: Basically most use cases we've errored out on. I think we need a better general purpose parser, perhaps building on something like this: http://pathvalidate.readthedocs.io/en/latest/pages/examples/sanitize.html#replace-symbols. But I also want to do some more research in the data/security community to see if there's a good all-around parser, especially when dealing with UTF8 and international symbols.

etl-scripts/northstar_to_user_table.py
sheyd: Will modify and update this.

etl-scripts/northstar_to_user_table.py
sheyd: Yes, it will give SSL invalid file warning.

etl-scripts/northstar_to_user_table.py
mshmsh5000: Seems safe to start out with.

etl-scripts/thor_northstar_to_user_table.py
sheyd: I've run into string substitution errors on string DB env, it's inserting an extra single-quote and breaking the string query. I have the code stashed and will push up another PR and create a backlog ticket to clean this up later. To get member count out sooner, going to pivot and focus on MailChimp scraping code. Merging in.

Issue #321, opened 2017-03-27
Data Analyst able to view members' MailChimp subscription status in Quasar users table


jamjensen:
_TASK_
-------------------
#### Ideal Deadline
3/28/17

#### Description
_MailChimp subscription status fields need to be added to Quasar `users` table and ETL'd out using existing MailChimp ETL code._

As a Data Analyst, I want to be able to view members' Mailchimp subscription status in Quasar `users` table so that I can determine DoSomething.org's accurate member count. 

#### Why This Matters
- In order to provide the organization/board/external partners with an accurate member count, we need to first know the number of members that are currently subscribed to our Mailchimp list. In order to know the number of members that are currently subscribed to our Mailchimp list, we need to add every member's MailChimp subscription status to the Quasar `users` table. 
- It's part of our Q1 goals. 

#### Definition of Done 
**Given** I query the Quasar `users` table
**When** I query on the MailChimp Subscription Status field
**Then** I am able to view every member's Mailchimp Subscription status

#### Additional Information
Blocking #310 and #294.

jamjensen:
- SH: "anticipate data tomorrow afternoon."

sheyd:
I have the backfill script running, and will update there's data in the DB.

sheyd:
@mshmsh5000 @DFurnes Can I get both your brain-spaces on this ETL script: https://github.com/DoSomething/quasar/blob/ns-to-users-mailchimp/etl-scripts/mailchimp_subscriber_to_user_table.py? I'm running it now to import the Mailchimp subscribers to Quasar table. Specifically, the way the logic works is it builds up a huge array then does all the DB processing at once. I think this is less desireable for two reasons:

1. Doing a backfill like I'm doing now takes forever to get any useful data in the table, and if it fails via API timeout or anything else, it just dies. 
2. It can consume lots of RAM that never gets cleared. And while not particularly an issue for the ETL instances we use, still something that's just not good systems architecture. 

Caveat is that there's more inline DB processing, and we've seen MySQL shit the bed from too many open connections/files before, but this is 1 request every API response, which take about 500-3000ms, so not too worried about it. I think it might be worth a quick refactor to inline the DB insert per API call. Thoughts/opinions?

DFurnes:
Yeah I think doing incremental inserts into the database is a really good idea. 👍

DFurnes:
> Caveat is that there's more inline DB processing, and we've seen MySQL shit the bed from too many open connections/files before

[...]

Issue #320, opened 2017-03-26
Complete Environment Substitution for Northstar User Table ETL


sheyd:
_BUG_
-------------------
#### Current Behavior
Prod and Thor environment config for Northstar -> Users table ingestion is currently in two different 
files. 

#### Desired Behavior
Prod and Thor environments should be substitutable via command line arguments passed to same ETL scripts.

#### Why This Matters
Env substitution should be handled via command line flags.

#### Relevant Screenshots + Links
https://github.com/DoSomething/quasar/tree/ns-to-user-env-cleanup

PR #326, opened 2017-04-04
Northstar Mailchimp and More Northstar Fields

sheyd:
#### What's this PR do?
* Adds Mailchimp import to Northstar table. 
* Updates Mobile Commons and `last_aurthenticated_at` fields for ingestion.

https://github.com/DoSomething/quasar/blob/ns-to-users-mailchimp/etl-scripts/mailchimp_subscriber_to_user_table.py

Tested again thor_users table in QA.

#321, #294, #310.

### DIFF
--- a/etl-scripts/mailchimp_subscriber_to_user_table.py
+++ b/etl-scripts/mailchimp_subscriber_to_user_table.py
@@ -1,7 +1,7 @@
 import requests
 from requests.auth import HTTPBasicAuth
 import time
-import datetime
+from datetime import datetime
 import MySQLdb
 import MySQLdb.converters
 import sys
@@ -41,6 +41,7 @@ db = MySQLdb.connect(host=config.host,  # hostname
                      passwd=config.pw,  # password
                      use_unicode=True,  # Use unicode for queries.
                      charset='utf8',    # Use UTF8 character set for queries.
+                     ssl=ca_settings,   # Connect using SSL
                      conv=conv_dict)    # datatype conversions
 
 cur = db.cursor()
@@ -81,8 +82,8 @@ origin_time = time_now - backfill_time
 # over designated time window by paginating via "offset" parameter.
 # Grabs in batches of 1000 to be nice to API. Can modify based on query times.
 info = {'status': 'subscribed',
-        'since_timestamp_opt': datetime.datetime.utcfromtimestamp(origin_time).isoformat(),
-        'before_timestamp_opt': datetime.datetime.utcfromtimestamp(time_now).isoformat(),
+        'since_timestamp_opt': datetime.utcfromtimestamp(origin_time).isoformat(),
+        'before_timestamp_opt': datetime.utcfromtimestamp(time_now).isoformat(),
         'count': 1000, 'fields': 'members.email_address,members.timestamp_opt,members.status,members.stats,members.list_id,members.location',
         'offset': member_offset}
 # Initialize Empty List for Total Members
@@ -134,8 +135,8 @@ while (len(member_array['members'])) > 1:
                  \"{0}\")".format(member_offset))
     db.commit()
     info = {'status': 'subscribed',
-            'since_timestamp_opt': datetime.datetime.utcfromtimestamp(origin_time).isoformat(),
-            'before_timestamp_opt': datetime.datetime.utcfromtimestamp(time_now).isoformat(),
+            'since_timestamp_opt': datetime.utcfromtimestamp(origin_time).isoformat(),
+            'before_timestamp_opt': datetime.utcfromtimestamp(time_now).isoformat(),
             'count': 1000,
             'fields': 'members.email_address,members.timestamp_opt,members.status,members.stats,members.list_id,members.location',
             'offset': member_offset}
--- a/etl-scripts/mailchimp_subscriber_to_user_table.py
+++ b/etl-scripts/mailchimp_subscriber_to_user_table.py
@@ -10,61 +10,30 @@ import MySQLdb.converters
 import sys
 import config
 
-# Set Backfill Hours, default of from supplied argument
-if not sys.argv[1]:
-    backfill_hours = 2
-else:
-    backfill_hours = int(sys.argv[1])
+def isInt(s):
+    """Check if value is type int and return boolean result.
+    Source at http://stackoverflow.com/questions/1265665/python-check-if-a-string-represents-an-int-without-using-try-except
+    """
+    try:
+        int(s)
+        return True
+    except ValueError:
+        return False
 
 # This is to track time of script
 start = time.time()
-
-# Start time in Unixtime to use internally in ETL processing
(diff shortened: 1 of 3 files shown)


etl-scripts/mailchimp_subscriber_to_user_table.py
```diff
@@ -0,0 +1,154 @@
...
+from requests.auth import HTTPBasicAuth
+import time
+import datetime
+import MySQLdb
+import MySQLdb.converters
+import sys
+import config
+
+
+def isInt(s):
```
DFurnes: It'd be nice to move this into a "helpers" file somewhere so you can import it into all of your scripts without having to copy-and-paste it into each file.

etl-scripts/mailchimp_subscriber_to_user_table.py
```diff
@@ -0,0 +1,154 @@
...
+conv_dict = MySQLdb.converters.conversions.copy()
+conv_dict[246] = float
+conv_dict[3] = int
+
+# Setup DB Connnection
+# Moved from earlier in script to not
+# have open MySQL connection that can timeout if
+# API scraping takes a long time. Have seen timeouts happen.
+
+ca_settings = {'ca': '/home/quasar/rds-combined-ca-bundle.pem'}
```
DFurnes: This variable doesn't seem to be used. Can you add a comment explaining where if it is?

etl-scripts/mailchimp_subscriber_to_user_table.py
```diff
@@ -0,0 +1,154 @@
...
+                     user=config.user,  # username
+                     passwd=config.pw,  # password
+                     use_unicode=True,  # Use unicode for queries.
+                     charset='utf8',    # Use UTF8 character set for queries.
+                     conv=conv_dict)    # datatype conversions
+
+cur = db.cursor()
+
+if len(sys.argv) == 3:
+    if isInt(sys.argv[1]):
```
DFurnes: Not necessary for merge, but it'd be nice to think about using a library to handle some of the lower level details of parsing command line arguments. I only did a quick search but [click](http://click.pocoo.org/5/) looks pretty nice.

etl-scripts/mailchimp_subscriber_to_user_table.py
```diff
@@ -0,0 +1,154 @@
...
+# Set backfill to seconds
+backfill_time = backfill_hours * 3600
+# Set backfill origin time to time_now - backfill_time in seconds
+origin_time = time_now - backfill_time
+
+# MailChimp API v3.0 Parameters to Get Subscribed Users
+# over designated time window by paginating via "offset" parameter.
+# Grabs in batches of 1000 to be nice to API. Can modify based on query times.
+info = {'status': 'subscribed',
+        'since_timestamp_opt': datetime.datetime.utcfromtimestamp(origin_time).isoformat(),
```
DFurnes: It looks like this `utcfromtimestamp` is the only part of `datetime` being used, so it could be nice to import it directly (like you do with `HTTPBasicAuth`) so you don't have to reach in each time you need it.

etl-scripts/mailchimp_subscriber_to_user_table.py
```diff
@@ -0,0 +1,154 @@
...
+        'before_timestamp_opt': datetime.datetime.utcfromtimestamp(time_now).isoformat(),
+        'count': 1000, 'fields': 'members.email_address,members.timestamp_opt,members.status,members.stats,members.list_id,members.location',
+        'offset': member_offset}
+# Initialize Empty List for Total Members
+total_members = []
+
+
+# Get first batch of users
+r = requests.get('https://us4.api.mailchimp.com/3.0/lists/f2fab1dfd4/members',
```
DFurnes: Is it possible to use our [web scraper](https://github.com/DoSomething/quasar/blob/master/etl-scripts/QuasarWebScraper.py) class here?

etl-scripts/thor_mailchimp_subscriber_to_user_table.py
```diff
@@ -0,0 +1,154 @@
...
+                 auth=HTTPBasicAuth(un, pw), params=info)
+member_array = r.json()
+
+# Iterate over entire list set till no more members are returned.
+while (len(member_array['members'])) > 1:
+    total_members.append(member_array['members'])
+    for member in total_members:
+        print("Processing next page. Up to %s processed." % member_offset)
+        for submember in member:
+            cur.execute("SELECT northstar_id FROM quasar.thor_users \
```
DFurnes: Is this table name the only difference between this script and `mailchimp_subscriber_to_user_table.py` above? If so, maybe it's worth combining them and adding a parameter to switch between environments.

etl-scripts/mailchimp_subscriber_to_user_table.py
sheyd: Created https://github.com/DoSomething/quasar/issues/327 to track this and definitely want to centralize a lot of our repeated code.

etl-scripts/mailchimp_subscriber_to_user_table.py
sheyd: Totally agree, would love to use a better parser 3rd party library. There are a couple more out there as well that I've come across and are pretty slick. Ticketized here: https://github.com/DoSomething/quasar/issues/328

etl-scripts/mailchimp_subscriber_to_user_table.py
sheyd: Yes, but since this MailChimp code is time-limited in functionality and works, I didn't want to spend too much time on re-factoring.

etl-scripts/thor_mailchimp_subscriber_to_user_table.py
sheyd: Already ticket for that here: :https://github.com/DoSomething/quasar/issues/320. Basically, ran into some string parsing errors from MySQL when passing the envs as vars. Mostly around inserted single quotes, but didn't have time to chase things down the rabbit hole. I want to re-visit this this quarter.

etl-scripts/mailchimp_subscriber_to_user_table.py
sheyd: Ah, forgot the `SSL` setting for the MySQL block. Updated.

etl-scripts/mailchimp_subscriber_to_user_table.py
```diff
@@ -0,0 +1,155 @@
...
+    try:
+        int(s)
+        return True
+    except ValueError:
+        return False
+
+
+# This is to track time of script
+start = time.time()
+
```
mshmsh5000: Not a blocker: Somewhere up here, define `mc_increment = 1000` or similar so you don't have to hard-code 1000 in multiple places below. This makes it easier to tweak this number for performance.

PR #339, opened 2017-05-02
Quasar-Blink Consumer

sheyd:
#### What's this PR do?
Hot off the presses!!! 
1. @RobSpectre hack day code updated with relevant DB entries, some tasty docstrings, and even a new private method!
2. It connects to and handles consuming messages via dedicated Blink Quasar RabbitMQ queue.
3. Parses messages to insert c.io sub/unsub status with corresponding timestamps to Quasar DB.

Any file below, except the ones in `misc-files`. That's essentially repo scratch space now.

Tested locally on Rob Spectre, after adding env vars to Prod, will test further.

This is our first code we're connecting to and consuming queue messages. Our deadline is May 15 at the latest, so we should def think and document optimizations that make sense for v.Next.

#338 

#### Questions:
For @mshmsh5000 and @DFurnes, some internal thoughts/easy wins totally happy to do before merge, just trying to think of MVP:
1. Document private methods.
2. Should this be its own function: https://github.com/DoSomething/quasar/blob/blink-consumer/etl-scripts/QuasarQueue.py#L106-L109
3. Should I break out some of the general "helper" functions, like MySQL query and Error Logging into it's own library file and import, so we move in the direction of not copy/pasting? :)


sheyd:
Also, my first PR ever that's successfully fully linted.

etl-scripts/QuasarQueue.py
```diff
@@ -0,0 +1,227 @@
...
+                 amqp_uri=config.AMQP_URI,
+                 amqp_queue=config.AMQP_QUEUE,
+                 amqp_exchange=config.AMQP_EXCHANGE,
+                 mysql_host=config.MYSQL_HOST,
+                 mysql_port=config.MYSQL_PORT,
+                 mysql_user=config.MYSQL_USER,
+                 mysql_password=config.MYSQL_PASSWORD,
+                 mysql_database=config.MYSQL_DATABASE,
+                 mysql_table=config.MYSQL_TABLE):
+        """Setup MySQL and AMQP connections and credentials."""
```
DFurnes: Definitely a "vNext" thing, but it'd be nice to pull MySQL stuff out into a separate `QuasarDatabase` class and have this `QuasarQueue` class only focused on queue related concerns.

etl-scripts/QuasarQueue.py
```diff
@@ -0,0 +1,227 @@
...
+                                                        mysql_port,
+                                                        mysql_user,
+                                                        mysql_password,
+                                                        mysql_database)
+        self.mysql_cursor = self.mysql_connection.cursor()
+
+    def start(self):
+        """Kick off consumer process to ingest messages.
+
+        Stays active until killed by keyboard interrupt (Ctrl-c or equivalent).
```
DFurnes: Another "vNext" thing – this feels like something we'd want to run continuously as a daemon right? If so, probably makes sense to just continue listening until the process receives a kill signal.

etl-scripts/QuasarQueue.py
```diff
@@ -0,0 +1,227 @@
...
+        """Parse and run DB query.
+
+        On error, raise exception and log why.
+        """
+        try:
+            self.mysql_cursor.execute(query)
+            self.mysql_connection.commit()
+            return True
+        except MySQLdb.DatabaseError as e:
+            raise QuasarQueueException(e)
```
DFurnes: Nice!

misc-code/QuasarQueue-Furnes-Improvements.py
```diff
@@ -0,0 +1,49 @@
+import pika
+
+import json
+
+
+class BlinkQueue:
```
DFurnes: 👻 😄

mshmsh5000:
Looks good! Obviously this is a proof of concept, but is this class called `QuasarQueue` because you intend to extend this class to use cases beyond c.io? Or will it be more accurate to call it `CustomerIOQueue` or similar?

I imagine we'll have routing logic in the future that identifies the incoming event either by exchange or some metadata in the payload itself. In which cases there may be a `QuasarQueue` with more generic responsibilities, and `CustomerIOEvent` or similar.

etl-scripts/QuasarQueue.py
mshmsh5000: Oh, isn't that what this does already?

etl-scripts/QuasarQueue.py
mshmsh5000: +1

etl-scripts/QuasarQueue.py
sheyd: @DFurnes @mshmsh5000 yes, it does. Rob recommended using something like Supervisor on this to make it a running process continually (@DFurnes, that's what we use for NSMCD, right?). I left the default as KeyboardInterrupt, but I can look at other interrupts to see how they run. Essentially, this would likely be an infinite running Jenkins job, which, come to think of it, is likely not the best idea. I'll find another way to wrap this up before deploying it. Worst case, I'll push this live, and we can figure out the best packaging/runtime over the next few days and see how it plays with Jenkins.

sheyd:
@mshmsh5000 It's the only thing in Quasar handling queueing right now, but yes, point taken. I'll rename QuasarQueue to `CustomerIOQueue`, and the processing script `CustomerIOMessageProcessing`, or something along those lines.

sheyd:
@mshmsh5000 Renamed the files. If they past muster, anything else I should get in before merge and testing out live?

mshmsh5000:
All 👍 from me. I imagined we'd do file & class renaming in the future anyhoo.

etl-scripts/QuasarQueue.py
DFurnes: @sheyd Yeah, we use supervisor for NSMCD! So far it hasn't let us down. 😄 But also yeah, this is totally what this does, I just wasn't reading close enough.

PR #341, opened 2017-05-03
Bug Fixes for Quasar-C.IO Consumer

sheyd:
#### What's this PR do?
1) Added new functionality to return mysql query results as well as if successful or not.
2) Added internal _bare_str method to remove special characters from query/string results.
3) Updated conversion from unixtimestamp to ISO8601 date format.
File diffs below.
Tested in prod already.
@sergii-tkachenko provided a test mechanism for testing live account. There were a few bugs I ironed out that are in this fix branch.
#338

### DIFF
--- a/etl-scripts/QuasarCustomerIOQueue.py
+++ b/etl-scripts/QuasarCustomerIOQueue.py
@@ -1,7 +1,9 @@
+from datetime import datetime
 import json
 import logging
 import MySQLdb
 from queue import Queue
+import re
 import time
 
 import pika
@@ -102,13 +104,15 @@ class QuasarQueue:
         message_type = self._message_type(message_data)
 
         if message_type:
-            email_address = ['data']['data']['email_address']
+            email_address = message_data['data']['data']['email_address']
             northstar_id = self.mysql_query("SELECT northstar_id "
                                             "FROM {1} WHERE email = \"{0}\";"
                                             "".format(email_address,
                                                       self.mysql_table))
-            if northstar_id:
-                query_results = self.insert_record(message_data, northstar_id,
+            if northstar_id[0]:
+                query_results = self.insert_record(message_data,
+                                                   self._bare_str(
+                                                       northstar_id[1]),
                                                    message_type)
             else:
                 pass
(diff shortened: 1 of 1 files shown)


etl-scripts/QuasarCustomerIOQueue.py
```diff
@@ -102,13 +104,15 @@ def _process_message(self, method_frame, message_data):
         message_type = self._message_type(message_data)
 
         if message_type:
-            email_address = ['data']['data']['email_address']
+            email_address = message_data['data']['data']['email_address']
```
DFurnes: Data, data, data!

etl-scripts/QuasarCustomerIOQueue.py
```diff
@@ -184,6 +188,12 @@ def _message_type(self, message_data):
         else:
             return False
 
+    def _bare_str(self, base_value):
```
DFurnes: This would make sense to extract as part of a "helpers" file in the future.

etl-scripts/QuasarCustomerIOQueue.py
```diff
@@ -192,7 +202,8 @@ def mysql_query(self, query):
         try:
             self.mysql_cursor.execute(query)
             self.mysql_connection.commit()
-            return True
+            results = self.mysql_cursor.fetchall()
+            return True, results
```
DFurnes: Is this a standard pattern in Python?

etl-scripts/QuasarCustomerIOQueue.py
sheyd: Was thinking the same exact thing. Def prime refactoring for helper library that will have All-The-Good-Things™ later this quarter, early Q3.

etl-scripts/QuasarCustomerIOQueue.py
sheyd: Standard as far as Stack Overflow is concerned, from the few examples I saw.

DFurnes: Okay! Definitely doesn't block merging, but seems a little odd at first glance since the first value is always `True` yeah?

sheyd: Yeah, it does, and only False on second. Good point, will have to ping Rob about proper syntax here (or just do more scanning of Python standards).

Issue #346, opened 2017-05-09
Invalid Message derails Quasar ingestion of c.io data


jamjensen:
_BUG_
-------------------
#### Current Behavior
- When using blink as a queuing mechanism to ingest c.io data into quasar, a specific type of invalid message prevents quasar from ingesting the queue. 

#### Desired Behavior
- This type of invalid message should not derail ingestion of c.io data to quasar but rather the process should be able to continue .

sheyd:
This was fixed in PR #350 hotfix. Closing.

Issue #344, opened 2017-05-08
Data Analyst able to view c.io status for members added to mailchimp between date of csv export (May 2) and May 15


jamjensen:
_TASK_
-------------------
#### Description
As a Data Analyst I want to be able to view the c.io subscription status of the members who were added to our mailchimp list after the date we exported the csv (May 2nd) until May 15th (date we go live with c.io) so that we can have an accurate understanding of the number of people that are active on our mailing list. 

#### Why This is Important
So that we can have an accurate member count. 

#### Definition of Done
**Given** Sena transfers c.io subscription status from mailchimp
**When** We query quasar user for c.io subscription status 
**Then** We're able to see c.io subscription status for members added during the time period listed above. 

#### Additional Information 
- expect 2-5k new accounts created

sheyd:
@jamjensen Just ran the updates this morning. Can you query against c.io subscription status in users table and let me know if you see any glaring disparities between that count and MailChimp counts we currently use as sources in Looker?

jamjensen:
After reviewing with @nmccormick619 we think there's 3 issues affecting validity of user subscription status: 

(1) 72,866 members with mailchimp_subscription_status = subscribed AND c.io_subscription_status = unsubscribed. 

- Number of members with mailchimp_subscription_status = subscribed: 3,007,268
- Number of members with customer_io_subscription_status = subscribed: 2,934,402

@sheyd I spot checked several of the users with differing mailchimp and c.io subscription statuses in mailchimp and for each one they appeared as subscribed to the `International` list and unsubscribed to the `Do Something Members` list.  

(2) We're seeing `unsubscribed` = FALSE for every single user. 

(3) C.io subscription status is unsubscribed AND mailchimp subscription status is NULL for ~700k users  
- Query:
 ```select u.email, u.customer_io_subscription_status, u.mailchimp_subscription_status, c.unsubscribed
from customer_io_export c
left join users u on u.email=c.email
where c.unsubscribed=FALSE
and u.customer_io_subscription_status='unsubscribed'
and u.mailchimp_subscription_status IS NULL
and u.email IS NOT NULL ```

sheyd:
Point 2 is fixed. Re-ran the update status on the `unsubscribed` columns based on the original MailChimp list imports.

sheyd:
Point #3 is *not* a valid comparison. The users table and customer_io_export table are not directly comparable. The users table is up to date every hour or so, whereas every info piece of customer_io_export is from export last week.

[...]

PR #347, opened 2017-05-12
Hotfix for Customer IO Ingestion Process.

sheyd:
#### What's this PR do?
* Moved messaging ack loop to run only if valid message type.
* Changed check for returned MySQL query status, but for
presence of actual value, since empty set will return True.

https://github.com/DoSomething/quasar/compare/customer-io-export-fix?expand=1#diff-23c3dc10a5677916cafbf3fad988148cL112

🤔 
There's an issue where if a new user has a "subscribed" status sent, the check for if `northstar_id` exists or not would always return True and parse/ack the message with no actual record update.

There's also an error out on checking for `query_results` in this line https://github.com/DoSomething/quasar/compare/customer-io-export-fix?expand=1#diff-23c3dc10a5677916cafbf3fad988148cL124 that would cause invalid message times to error out the script.

#344 , #346

### DIFF
--- a/etl-scripts/QuasarCustomerIOQueue.py
+++ b/etl-scripts/QuasarCustomerIOQueue.py
@@ -109,27 +109,24 @@ class QuasarQueue:
                                             "FROM {1} WHERE email = \"{0}\";"
                                             "".format(email_address,
                                                       self.mysql_table))
-            if northstar_id[0]:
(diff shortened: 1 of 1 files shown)


sheyd:
@mshmsh5000 @DFurnes Totally ignore this PR for now. Testing the code for more de-bugging. Current version doesn't work.

PR #350, opened 2017-05-14
Quasar c.io queue processing hotfix to backup unparsed messages to DB.

sheyd:
#### What's this PR do?
1. Changes Northstar ID parsing exception to not be requeued butto put relevant data (email, c-io-status, c-io-timestamp) into a Quasar queue backup table.
2. Updates main queue processor to use new derived class, leaving
original code untouched to revisit this sprint or next.

This is a hotfix to be able to parse through the original 3.6 million Customer.io CSV imports before Monday without having the queue processing job exit regularly.

Two files below.
Tested in Prod and verified.
The original QuasarQueue Process used internal queueing mechanisms to store retried messages, but would stall with errors that couldn't be de-bugged by deadline. This workaround provides a persistent backup mechanism of the pertinent data we can backfill later this sprint or next sprint, while making sure we have the bulk of the data ingested.

#344 #346

### DIFF
--- a/etl-scripts/QuasarCustomerIOProcessQueue.py
+++ b/etl-scripts/QuasarCustomerIOProcessQueue.py
@@ -1,4 +1,4 @@
-from QuasarCustomerIOQueue import QuasarQueue
+from QuasarCustomerIOQueueDBBackup import QuasarQueue
 import logging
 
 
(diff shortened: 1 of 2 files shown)


sheyd:
@DFurnes @mshmsh5000 I'm merging this code in as it's tested in prod and it's a hotfix, but wanted, as always 👁‍🗨  on this in case anything better/obvious can be documented for improvement.

Issue #351, opened 2017-05-16
Write Re-Publish Code for Queue Re-insertion and Error'd Objects


sheyd:
_TASK_
-------------------
#### Description
As a Data engineer I want to requeue messages that lack Northstar IDs until NS IDs are imported into the `users` table so that we're able to update c.io subscription status. 

#### Why This is Important
Requeing messages that lack Northstar IDs is critical because we're only able to update c.io subscription if a record has a Northstar ID. 

#### Definition of Done
**Given**
**When**
**Then**

sheyd:
Using Pika library `basic_nack` on consumer instead of manually re-parsing/publishing message works in so far as new messages with no Northstar ID get requeued. The problem is, there's no TTL, and the messages are instantly re-processed and requeued, to tune of 9 unacked messages has 600-800 Requests per second back the to RabbitMQ server. I'll need to figure out how to put a reasonable TTL on this, or if not available as a param via `basic_nack`, work out publish code.

sheyd:
PR merged, and has been successfully running in Prod for a few hours now.

PR #353, opened 2017-05-22
Requeue Unparsable Objects

sheyd:
#### What's this PR do?
* Updates C.IO ingestion code to requeue and exit on objects that don't have a Northstar ID.
The 2 files below.
Tested in prod. 
*^I'm really starting to hate this statement.^*
The backup table hotfix we had is a one off, and not very maintainable. To keep things at their simplest, we're adding a Jenkins pipeline dependency so job exits when hits new ideas, and re-runs after next Northstar ID ingestion.
Also, long-running Jenkins jobs aren't really sustainable.
#351


sheyd:
Every 10 minutes.

PR #364, opened 2017-06-16
Add README template and setup instructions

xianny:
#### What's this PR do?

* Put more details in the README to make it a one-stop-shop for project info
* Describe setup steps for Quasar local development (Python only)

Just compare changes

NA

Set up instructions here are just for setting up an isolated Python dev environment. Looked into provisioning a Vagrant box but I think it makes more sense to do it when I have a better idea of how we are using Ansible and other tools. 

NA

#### Questions:

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,9 +90,9 @@ None yet. Let's add automated tests!
 
 ## Contributing
 
-[CONTRIBUTING.md](CONTRIBUTING.md) // add process notes?
-[Pull request template](PULL_REQUEST_TEMPLATE)
-[Issue template](issue_template.md)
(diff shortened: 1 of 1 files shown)


xianny:
Virtualenv is totally fine! Virtualenv is standard for python2 but pyenv is
recommended for python3 (meant to replace virtualenv). I run both. My
suggestion is to keep pyenv instructions for new devs, but anyone who
already has an environment management setup should not feel compelled to
switch.

mshmsh5000:
+1! Makes total sense to get new users on `pyenv`, especially ahead of new people joining the team this summer.

Issue #363, opened 2017-06-15
6/15 Rogue/Data Meeting Recap


ngjo:
**6/15 Rogue/Data Meeting Recap**

Some areas of discrepancy and next steps:
1. SMS Duplicate issue on Gambit for folks who messaged in w/ +160 characters creating multiple signups and multiple reportbacks came up again (example: https://www.dosomething.org/us/user/5007132).  This is a problem because we're over reporting  signups and reportbacks. (Issue first flagged here: https://trello.com/c/8WAc2icF/1657-duplicate-rbs-w-sms-as-the-source) 
2. We came across a user (Northstar ID: 5581ca74469c6475138b5518, UID: 2806506) who has reportbacks for a campaign, but doesn't have a signup for that campaign. This is a problem because Rogue is looking for signups. (A larger question: do we want to count reportbacks if the signup is deleted?)
3. Quasar should be ingesting `signup_id` instead of `signup_event_id` as that is the signup for a member for a given campaign. This will allow Looker to look at the `signup_id` or signup count.
4. It seems that SMS Betas are being assigned a NULL value for their UIDs - they are not being counted as a signup (I think?)
5. In Looker, for pre-Gambit campaigns, we will show the current dashboards and won't try to hook it up with Rogue data. For post-Gambit campaigns, we will show dashboards w/ Rogue data as Gambit receives signups/reportbacks which are sent to Phoenix that are in turn sent to Rogue. We made this decision because the main discrepancy for signups on the pre-Gambit campaigns are due to the de-duped Mobile Commons data. Working with this query is difficult, and we felt, unnecessary if pre-Gambit campaigns hold historic data. Here's an example of pre-Gambit, Power to the Period data: 

![fullsizerender 1](https://user-images.githubusercontent.com/2658867/27193814-4c7805e6-51ce-11e7-9b2e-c8323abfc202.jpg)

sbsmith86:
There was one campaign where there was a discrepancy between Rogue Looker data and Rogue Prod data
![img_7859](https://user-images.githubusercontent.com/1700409/27195178-b062788e-51d3-11e7-8bfe-9f00d9fbf061.JPG)

sbsmith86:
Documenting the Rogue queries that we ran against prod:

```
/* total signups: 3,407,695 */
SELECT COUNT(*) FROM `signups`;

/* [7249] captain obvious signups: 5107        DISTINCT: 5,107 */
[...]

PR #366, opened 2017-06-19
Use column name `signup_id`, instead of `signup_event_id`

sheyd:
#### What's this PR do?
Data Warehouse `campaign_activity` column updated to use `signup_id` instead of `signup_event_id`.
Two lines of code changed.
Unproudly tested in Prod.
#363


sheyd:
@mshmsh5000 Gah, meant #363! Thanks for catch.

Issue #374, opened 2017-06-27
Backfill Historical Northstar Records to Update Last_Accessed / Logged_in


jamjensen:
_TASK_
-------------------
#### Description
As a Data Analyst I want all updated records from Northstar in the Quasar.Users table so that I can see the most up-to-date user info and the most recent time all users have accessed the site. 

#### Why This is Important
Backfilling all Northstar records is necessary so that we have the most up-to-date user information for all user in Quasar.

#### Definition of Done
**Given** Sena completes the backfill of Northstar historical records
**When** James queries the quasar users table 
**Then**  He should see updated records from early April 2017, when we started to collect data in Northstar. 

#### Additional Information
- Currently, we ingest user information when a user creates an account. However, we do not ingest user information from Northstar when they update their profile. To be fixed in #373. 

- Ask Dave if we did a bulk migration of last_accessed info from phoenix db into Northstar.

DFurnes:
> Ask Dave if we did a bulk migration of last_accessed info from phoenix db into Northstar.

We did not.

sheyd:
Backfill kicked off. Will update when it's done.

jamjensen:
@sheyd can you include a link to the backfill? I can't find in among the Jenkins jobs. 

I've also noticed that there are 300k (and counting) fewer users with `customer_io_subscription_status` as subscribed since I last checked this morning. Can you please see if the backfill is overwriting subscriptions statuses, and if so, pause it?

jamjensen:
**Update** 
While backfilling Northstar records to the Quasar Users table we accidentally set c.io subscription status values to NULL. Documented here -- https://dosomething.slack.com/archives/C0FP51V2T/p1498616931083803 

@sheyd can you provide an update of where we are with updating all users c.io subscription status in the users table based off of the latest c.io export?

sheyd:
@jamjensen I've confirmed CSV numbers match up with c.io web numbers. I've imported the CSV into `c_io_export_fix_missing` table, and am doing some data validation. I'm not worried too much about the sub/unsub status updates, but a lot of the sub/unsub times are epoch 00:00:00 times. Looking into why this is.

sheyd:
@jamjensen Confirmed that there's only 175,556 email addresses both in CSV export as well in the c.io Web UI that have a . Didn't catch this before, but looks like we didn't have a whole lot of subscribe times in c.io to begin with, and need to re-fill the times from the Mailchimp sub archive tables. 

[...]

PR #379, opened 2017-06-29
Make Record Updates from Northstar to Users Table non-destructive

sheyd:
#### What's this PR do?
Updated the Northstar->Users ETL script to use INSERT...ON DUPLICATE KEY UPDATE syntax to not overwrite pre-existing fields of data.
One file below.
Already tested on test table in Prod environment.
We previously used `REPLACE` which on a backfill update, nuked about 1M records fields we backfill from other ETL jobs to the same table.
#374.

### DIFF
--- a/etl-scripts/northstar_to_user_table.py
+++ b/etl-scripts/northstar_to_user_table.py
@@ -105,7 +105,7 @@ def to_string(base_value):
 while nextPage is True:
     current_page = ns_fetcher.getUsers(100, i)
     for user in current_page:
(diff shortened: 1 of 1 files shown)


DFurnes:
I'd love to prioritize extracting a `Database` class to handle the logic of updating rows in the warehouse so we don't have to write potentially error-prone queries like this by hand.

sheyd:
We can think about prioritizing using an ORM type thing like SQLAlchemy (https://www.sqlalchemy.org/) to provide some protection against this type of thing. Would also potentially make migrating to different DB backend should we ever need to easier.

Issue #373, opened 2017-06-27
Update Quasar Users Table to Handle Updated Records from Northstar


jamjensen:
_TASK_
-------------------
#### Description
As a data engineer, I want to extend the functionality of the Northstar-to-Quasar user ingestion ETL script so that it ingests updated records (in a given backfill time-period) from Northstar; such as,`last_accessed`, `last_logged_in`, and user profile fields, e.g.`phone_number`.

#### Why This is Important
This is necessary for us to (1) analyze user behavior on site; (2) have the most up-to-date user records; (3) remove reliance on phoenix database for user data. 

#### Definition of Done
**Given** Northstar API endpoint is fully exposed
**When** Sena confirms that the extended ETL script pulls updated Northstar records based on a given backfill time-period
**Then** We (team Quasar) will perpetually ingest updated Northstar records into Quasar

jamjensen:
@sheyd are you still blocked by #361?

DFurnes:
@jamjensen That pull request was deployed to production earlier today! 🚀

sheyd:
I'm not blocked by #361 anymore. I've updated the DS Northstar scraper code to call the updated NS API endpoints. I'm updating the ETL process to parse and update the DB.

sheyd:
Got the DS Northstar Scraper updated to handle new endpoints. Finishing updated ETL code.

jamjensen:
@sheyd were you able to finish updating Northstar-to-Quasar user ingestion ETL script?

sheyd:
Code has been merged, and I've updated the Jenkins Northstar -> Users table ingestion to use the new backfill process. It pulls updated and newly created members based on an hourly backfill every 15 minutes (up from 5). Not the best setup as, there is overlap 3-4 times an hour, and it lowers our 5 minute updates, but I don't want to tax the Northstar API too much, and our 5 minute runs were getting a bit onerous. 

I've done a test run and have seen positive results in the DB. Assigning to @jamjensen for further confirmation throughout the day that we're seeing updated `last_accessed` and `last_logged_in` fields to confirm updates from backfill, as well as `northstar_created_at_timestamp` fields that are propagating from today's date(7/13/17), ensuring we have full data fidelity. 

From the code side and ETL side, this work is complete, unless erroneous data is coming through.

sheyd:
[...]

PR #385, opened 2017-07-12
Ns to user updated since

sheyd:
#### What's this PR do?
Adds a backfill script to get diffs from Northstar for created_at and updated_at dates for Quasar user table.
Two files below.
Tested on small sample set backfill.
So far we've only been able to do full backfills, this will provide us with on-going updates on a much smaller time delay.
#373

### DIFF
--- a/etl-scripts/DSNorthstarScraper.py
+++ b/etl-scripts/DSNorthstarScraper.py
@@ -1,8 +1,10 @@
 import json
-import config
+
 import oauthlib
 from oauthlib.oauth2 import BackendApplicationClient
 from requests_oauthlib import OAuth2Session
+
+import config
 from QuasarWebScraper import Scraper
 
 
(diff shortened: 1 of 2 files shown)


etl-scripts/DSNorthstarScraper.py
```diff
@@ -49,37 +54,101 @@ def post(self, path, body=[]):
...
+
+    def getUsersCreatedSince(self, created_at, users=100, page_number=1):
+        """Get users in max batch size based on created date from Norsthar API.
+
+        Args:
+            users (int): Users per page returned, default 100, max 100.
+            page_number (int): Page number to return, default 1.
+            created_at (str): Created at parameter to backfill since.
+        """
+        params = ('?page=' + str(page_number) + '&limit=' + str(users) +
```
xianny: How about using a dictionary instead of string concatenation?

sheyd: That doesn't work for certain URL encodings, and as mentioned here: https://stackoverflow.com/questions/23496750/how-to-prevent-python-requests-from-percent-encoding-my-urls

Basically in this case `&after[updated_at]=` gets munged up by percent encoding. Happy to consider cleaner workarounds. So far, we haven't had too many URI's like this, so the SO solution above of over-riding the request sessions code hasn't been warranted, and seems like a hacky solution at best.

Issue #396, opened 2017-07-25
Bug in Campaign Activity Backfill Script


sheyd:
_BUG_
-------------------
#### Current Behavior
Campaign activity backfill wasn't inserting records into the Quasar DB.

#### Desired Behavior
Campaign activity backfill to properly update all user activity records.
#### Steps to Replicate
Run the Campaign Activity Backfill Jenkins job and query the Quasar DB for updated records via:
```
select count(distinct signup_id)
from campaign_activity
where `signup_created_at` > '2017-07-24' 
```
#### Why This Matters
Accurate member activity counts for broadcasts, dashboards, and org.

sheyd:
There's a small logic check bug in Quasar package 0.1.0 that requires checking for the proper pagination point for backfill vs full_backfill. Updating the code now.

sheyd:
Fixed and tested in Prod.

PR #397, opened 2017-07-25
Bugfix for Rogue Backfill Diff Not Updating Records

sheyd:
#### What's this PR do?
1. Added current_page =1 assignment when doing non-full backfill.
2. Full backfill from the top is state dependent on setting
Quasar DB status tracking page to 1, not entirely happy about
this, but we don't usually do full backfills, and it's a known
quantity.
One file below.
Test locally and records will be updated.
#396

### DIFF
--- a/quasar/campaign_activity.py
+++ b/quasar/campaign_activity.py
@@ -14,10 +14,11 @@ campaign_activity_table = config.CAMPAIGN_ACTIVITY_TABLE
 
 
 def full_backfill():
+    """Processes Rogue data starting from last processed page recorded in 
+    ROGUE_PROGRESS_TABLE; to restart from beginning, page_progress has to
+    manually be set to 1.
+    """
     _backfill()
-    # There is an implicit state dependency here that requires
-    # setting Quasar DB page_progress to 1 for full backfill restart.
-
 
 def backfill_since():
     _backfill(hours=sys.argv[1])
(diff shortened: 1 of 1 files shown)


quasar/campaign_activity.py
```diff
@@ -23,17 +25,19 @@ def backfill_since():
...
                    headers={'X-DS-Rogue-API-Key': config.DS_ROGUE_API_KEY},
                    params={'page': 1, 'limit': 40, 'filter[updated_at]': start_time})
     final_page = rogueAPI.get('')['meta']['pagination']['total_pages']
-    current_page = _get_start_page(db)
+
+    if hours is not None:
+        print("Current backfill hours are %s." % hours)
+        current_page = 1
+    else:
+        current_page = _get_start_page(db)
```
xianny: Not sure I understand what's going on here. 

Sounds like:

If we are backfilling x hours, set current page to 1

If we're doing full backfill, then get the start page from the db.

Is that right?

quasar/campaign_activity.py
```diff
@@ -15,6 +15,8 @@
 
 def full_backfill():
     _backfill()
+    # There is an implicit state dependency here that requires
+    # setting Quasar DB page_progress to 1 for full backfill restart.
```
xianny: I'm a bit confused by this comment here. Could this be a docstring? E.g. "Processes Rogue data starting from last processed page recorded in ROGUE_PROGRESS_TABLE; to restart, we have to manually set page_progress to 1".

quasar/campaign_activity.py
sheyd: @xianny Yes, that's exactly what's going on!

quasar/campaign_activity.py
sheyd: You're right, I'll update this.

PR #405, opened 2017-08-07
Removal Ansible and Add Dev Env Import Check

sheyd:
#### What's this PR do?
1. Removing extraneous ansible configs that are now in tower repo.
2. Added Dev specific file import for dev_priv, since fails if file
doesn't exist.
One file change below and code removal.
Checkout and run file locally.
Last merged in branch had a `dev_priv` dependency that doesn't exist on Prod/Stage servers, and caused code to fail unless dummy file was created.


quasar/config/__init__.py
```diff
@@ -1,10 +1,13 @@
 import os
 import sys
 
-from . import dev_priv as config
+from . import dev as config
 
 env = os.environ['ENV']
 
+if env == "DEV":
+    from . import dev_priv as config
```
sheyd: @xianny Hmmm, base on this, what about removing line 4 of the file, and just renaming `dev.py` to something like `dev_template.py` and updating docs?

xianny: Let's keep your L4 change (import dev instead of dev_priv) but remove L8-9. I think it's better to have either "default env" or "staging/prod". Having "dev", "other kind of dev", "staging/prod"  could be a bit confusing. Updating docs is a good idea - maybe we can add to the *Development* section to ask for a working *dev.py* if needed?

sheyd: So, basically, (just re-cap to avoid confusion):
1. Leave in dev.py, and keep L4 change, removing `dev_priv` specification. 
2. Add documentation that for functional dev.py config.
3. We'd have to add `dev.py` to `.gitignore`, so we don't have to keep using `dev_priv.py` locally and commenting/uncommenting it everywhere, while being able to alter local `dev.py` at will? If not the case, I'd prefer to use some contingency where we can set an env variable and local dev immediately. Or some similar mechanism.

xianny: Yes, that sounds good. Let's keep a copy of dev.py in the GH repo but add it to .gitignore so there won't be updates.

PR #408, opened 2017-08-08
Updated cursor reference based on DB Class object.

sheyd:
#### What's this PR do?
Bugfix for cursor reference in legacy jc_subscribers.py code.

One file below.
Test locally.


sheyd:
@xianny Tagged you for reference, merging in to fix prod bug.

PR #410, opened 2017-08-08
Only fetch oauth token if needed

xianny:
#### What's this PR do?
Stop fetching oauth token with every request. Fetch once when connection is first initialised, and then only re-fetch if we encounter auth errors.

Look at DSNorthstarScraper.py file

We can use python REPL. Try this code snippet:

```
import quasar
import quasar.DSNorthstarScraper
s = quasar.DSNorthstarScraper.NorthstarScraper('https://northstar.dosomething.org')
r = s.get('/v1/users', {'limit':5})
r.status_code # should be 200
s.auth_headers['Authorization'] = s.auth_headers['Authorization'] + 'blah'
r = s.get('/v1/users', {'limit':5})
r.status_code # should be 200 still
```


xianny:
@sheyd added another commit - refactored API scraping code everywhere.

sheyd:
Haha, interesting bug (feature?), you can approve PR's multiple times via the `Files Changed` tab.

PR #413, opened 2017-08-22
Gladiator import

sheyd:
#### What's this PR do?
Adds code to pull in campaign data from Gladiator.
Files below.
Checkout local, get schema dump, and hit API!


sheyd:
@xianny Pushing this live, since confirmed on prod, but please still comment if you have any thoughts :)

PR #426, opened 2017-09-25
Moco message v2

sheyd:
#### What's this PR do?
* Updates SQL schema to latest prod version for dev.
* Bumps Quasar version to `0.1.2`
* Refactors MoCo Messaging Scraper code to track last status.

https://github.com/DoSomething/quasar/compare/moco-message-v2?expand=1#diff-223272725de440c15902f7fe1e78b362 is most important place.

Import latest schema to local dev, and run `scrape_moco_messages` with flushed out values in DB tables.

### DIFF
--- a/data/sql/migrations/000_schema_quasar.sql
+++ b/data/sql/migrations/000_schema_quasar.sql
@@ -5,9 +5,9 @@
 # http://www.sequelpro.com/
 # https://github.com/sequelpro/sequelpro
 #
-# Host: quasar-staging.c9ajz690mens.us-east-1.rds.amazonaws.com (MySQL 5.7.17-log)
+# Host: quasar.c9ajz690mens.us-east-1.rds.amazonaws.com (MySQL 5.7.17-log)
 # Database: quasar
-# Generation Time: 2017-07-11 17:31:21 +0000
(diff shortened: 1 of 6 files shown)


data/sql/migrations/000_schema_quasar.sql
```diff
@@ -219,18 +192,29 @@ CREATE TABLE `users` (
...
 
-CREATE TABLE `users_test` (
+DROP TABLE IF EXISTS `users_log`;
+
+CREATE TABLE `users_log` (
   `northstar_id` varchar(26) COLLATE utf8mb4_unicode_ci NOT NULL,
   `northstar_created_at_timestamp` datetime DEFAULT NULL,
-  `last_logged_in` datetime DEFAULT NULL,
-  `last_accessed` datetime DEFAULT NULL,
+  `last_logged_in` datetime NOT NULL,
```
sohaibhasan: I'm not familiar enough with MySQL to know what happens when you change something from DEFAULT NULL to NOT NULL. If it was DEFAULT NULL before, why did we see all those 0000-00-00 00:00:00 values? Or was that how we were receiving those values from northstar?

PR #427, opened 2017-09-29
Updated some type and tuple reference errors.

sheyd:
#### What's this PR do?
Fixes 1 logic parsing error in the code (`==` instead of `!=`) and updates some old references from refactor that used str type instead of int for tuple references.

One file below.

Check out to local, build, and scrape_moco_messages.


sheyd:
@sohaibhasan Just tagging you on this as best practice. Fix has been tested successfully. Pushing live.

PR #432, opened 2017-10-04
Queue abstract

sheyd:
#### What's this PR do?
Adds abstracted QuasarQueue processing class to prep for ingesting multiple different queue messages. It's designed so the `on_message` function will be over-ridden by child classes to perform whatever operations we need either per queue, or per queue type message.
Added file below.
Checkout branch, and if you have appropriate Blink Queue credentials, setup test python class with single instance of QuasarQueue, then run queue.start_consume. You should see one queued message and its contents every half second.

All of our queue logic has traditionally been in the customerio.py file, but it has tightly coupled logic to only handle certain queue types.


quasar/queue.py
```diff
@@ -0,0 +1,93 @@
...
+    def on_message(self, channel, method_frame, header_frame,
+                   body, output="true"):
+        message_data = self.body_decode(body)
+        if output == "true":
+            print("Message received.")
+            print("Message method_frame is {}".format(method_frame))
+            print("Message header_frame is {}".format(header_frame))
+            print("Message body is {}".format(body))
+        print("Republishing message.")
+        self.pub_message(message_data)
```
DFurnes: Is it necessary to "pop" a message off the queue and then re-publish it here? Seems a little dangerous!

quasar/queue.py
```diff
@@ -0,0 +1,93 @@
...
+            self.channel.start_consuming()
+            logging.info("{0} consumer started.".format(self.amqp_queue))
+        except KeyboardInterrupt:
+            self.stop_consume()
+        self.connection.close()
+
+    def stop_consume(self, tag="quasar_consumer"):
+        self.channel.basic_cancel(tag)
+
+    def on_message(self, channel, method_frame, header_frame,
```
DFurnes: Instead of subclasses overriding this `on_message`, I think it might be helpful to have a `process` method that gets called from here. That way all the "machinery" happening in `on_message` stays intact.

```py
class Queue
  def on_message(self, channel, method_frame, header_frame, body, output="true"):
    # all the on_message things....
    self.process(message_data);
    # finish all the on_message things...

 def process(message_data)
    # stub!

class CustomerIoQueue
  def process(message_data)
     # the specific things for Customer.io queue processing!
```

quasar/queue.py
```diff
@@ -0,0 +1,93 @@
...
+            print("Message received.")
+            print("Message method_frame is {}".format(method_frame))
+            print("Message header_frame is {}".format(header_frame))
+            print("Message body is {}".format(body))
+        print("Republishing message.")
+        self.pub_message(message_data)
+        print("Acking message.")
+        self.ack_message(method_frame)
+        print("Original message ack'd.")
+        time.sleep(0.5)
```
DFurnes: Let's add a comment to explain that this is for rate-limiting. It may make sense in the future to add a "throughput" setting on the Queue class that tells it how many messages to grab per minute (and then calculates this "sleep" timer from that).

quasar/queue.py
```diff
@@ -0,0 +1,93 @@
...
+    def __init__(self,
+                 amqp_uri=config.AMQP_URI,
+                 amqp_queue=config.AMQP_QUEUE,
+                 amqp_exchange=config.AMQP_EXCHANGE):
+        """Setup MySQL and AMQP connections and credentials."""
+        self.params = pika.URLParameters(amqp_uri)
+        self.params.socket_timeout = 5
+        self.connection = pika.BlockingConnection(self.params)
+        self.channel = self.connection.channel()
+        self.channel.basic_qos(prefetch_count=100)
```
DFurnes: From our IRL conversation about acking & republishing – what happens here if we prefetch 100 messages off of the queue, but the queue listener crashes when processing them halfway through?

sheyd:
@DFurnes Just to keep things simple for now, I've added separate test methods, moved to using `basic_nack` instead of repub/ack which will requeue the message sanely.

Only further thought I had to ship this out was to maybe have the test methods be a separate class? Like TestQuasar, which instantiates the QuasarQueue class, and overloads the methods. I see benefits to both approaches. Let me know which one you think is best, and I'll modify, and then it's ready for 🚢 .

DFurnes:
It depends how you'll use them! I think it's fine to keep them on the base queue for now, since you'll already be extending this class for different queues and having to swap out an `extends TestQueue` may just add complexity.

PR #442, opened 2017-10-19
Updates queries to NULL insert on no location data.

sheyd:
#### What's this PR do?
If no location data exists in a profile, it ensures the query to the relevant `loc_*` DB fields are NULL'd.
One file below.
Checkout branch, pull down some MoCo Profiles locally, and iterate over the test set.
I hate putting the if check clause in the `_import_record` function, since I think it's cleaner to add blank profile fields into the original array, but it's less efficient, and also I was getting lost in subkey dict rabbit hole syntax.
https://www.pivotaltracker.com/story/show/151412624

### DIFF
--- a/quasar/moco_xml_to_quasar.py
+++ b/quasar/moco_xml_to_quasar.py
@@ -1,6 +1,4 @@
-import json
 import os
-import sys
 
 import boto3
 import xmltodict
--- a/quasar/moco_xml_to_quasar.py
+++ b/quasar/moco_xml_to_quasar.py
@@ -1,4 +1,6 @@
+import json
 import os
+import sys
 
 import boto3
 import xmltodict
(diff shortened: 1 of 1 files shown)


quasar/moco_xml_to_quasar.py
```diff
@@ -15,28 +15,48 @@
...
+                                   strip_str(profile['location']['state']),
+                                   profile['location']['postal_code'],
+                                   strip_str(profile['location']['country']),
+                                   profile['location']['latitude'],
+                                   profile['location']['longitude'])
+    else:
+        str = ''.join(("REPLACE INTO quasar.moco_profile_import ",
+                       "VALUES ({}, '{}', '{}', '{}', '{}', '{}', ",
+                       "'{}', '{}', '{}', '{}', '{}', '{}', '{}', NULL, ",
+                       "NULL, NULL, NULL, POINT(NULL, NULL))",
```
DFurnes: To make sure I'm following correctly: the reason we need this conditional is because `null` values for these fields aren't otherwise correctly saved as "real" nulls in the MySQL table unless we do this, right?

DFurnes: Can we try passing the bindings to [`MySQLCursor.execute`](https://dev.mysql.com/doc/connector-python/en/connector-python-api-mysqlcursor-execute.html) so it can fill them in, rather than doing this with plain string formatting?

sheyd: Talked to Dave about using bindings rather than string replacement for this. We're going to run into issues on this sooner rather than later. Creating ticket and will prioritize, especially as we swap over to PostgreSQL.

PR #444, opened 2017-10-27
Customer IO ETL Queue Refactoring

sheyd:
#### What's this PR do?
Hoo-boy, this is a big one, not code wise, but refactoring wise. Here we go:

* Every queue object now is written to an event table that captures all JSON events from c.io. It writes `meta` and `data` to two columns, and indexes on Northstar ID and C.IO Event ID. This is the `log_event` method.
* Every queue event object now is also written to a c.io specific customer event log table. This is handled by the `customer_event` method. 
* For parity with how we currently process c.io queue events, we update `users` table via the `legacy_sub_unsub` method, and if Northstar ID's don't exist yet (haven't been backfilled to `users` table), they get written to a backlog table, instead of being requeued.
* The `legacy_update_users` takes all backlog tables events, and if Northstar users are imported, updates their C.IO status, and removes the record from the backlog table, otherwise, the record stays in.

In absence of @DFurnes being out, tagging @mshmsh5000 for a 2nd set of 👀 
Files below. 
Create schema in Pivotal ticket locally, and modify code to re-queue messages after each test method.
We're doing this to capture all C.IO events and also allows us to completely empty our queues.
https://www.pivotaltracker.com/story/show/150851676


data/sql/migrations/000_schema_quasar.sql
```diff
@@ -194,7 +194,7 @@ CREATE TABLE `users` (
 
 DELIMITER ;;
 /*!50003 SET SESSION SQL_MODE="NO_ENGINE_SUBSTITUTION" */;;
-/*!50003 CREATE */ /*!50017 DEFINER=`quasaradm`@`%` */ /*!50003 TRIGGER `userlog` AFTER UPDATE ON `users` FOR EACH ROW BEGIN  
+/*!50003 CREATE */ /*!50017 DEFINER=`root`@`%` */ /*!50003 TRIGGER `userlog` AFTER UPDATE ON `users` FOR EACH ROW BEGIN
```
mshmsh5000: Any significance to the change in user?

sheyd:
@sohaibhasan Can you take one more look? I've added the `@staticmethod` call and lint cleaned up a few things as well, despite a small loss in efficiency by assigning a var ;)

quasar/cio_queue_process.py
```diff
@@ -0,0 +1,90 @@
...
+db = Database()
+
+
+class CioQueue(QuasarQueue):
+
+    def process_message(self, message_data):
+        print("Processing C.IO event id: {}.".format(message_data['data']['event_id']))
+        log_event(db, message_data)
+        customer_event(db, message_data)
+        if (message_data['data']['event_type'] == 'customer_subscribed' or
```
mshmsh5000: No biggie, but slightly more terse, still readable statement is

`if (message_data['data']['event_type'] in ['customer_subscribed', 'customer_unsubscribed'])`

data/sql/migrations/000_schema_quasar.sql
sheyd: @mshmsh5000 Yeah, on local dev, we don't have `quasaradm` user, so writing to `users_log` errors out. This fixes that.

quasar/cio_queue_process.py
```diff
@@ -44,28 +45,29 @@ def legacy_sub_unsub(db, message_data):
...
+                      strip_str(nsid)))
     else:
         db.query_str("INSERT INTO cio.legacy_sub_backlog VALUES (%s, %s, %s)",
                      (status,
                       u2i(message_data['data']['timestamp']),
                       message_data['data']['data']['customer_id']))
 
 
+@staticmethod
 def legacy_update_users():
```
sohaibhasan: It looks like this isn't in the class? The idea behind the staticmethod decorator is to include it in the class while essentially letting people know it is applicable outside of the class and doesn't depend on the class itself.

sheyd: It isn't... I thought it had context outside of that. Sorry. I'm going to remove the line, and we can re-visit. It's more important to get this code out right now. And I don't get what `@staticmethod` is or where it belongs enough to make this call.

sohaibhasan: All good. Fine to look more deeply into the pros and cons at a later date

PR #445, opened 2017-10-27
Remove ignore to investigate dup entries.

sheyd:
#### What's this PR do?
We noticed some duplicate objects in queue. We're recording everything for now to assess how often this happens.
One file below.
Doing it live. On a Friday. Like all good ideas.


sheyd:
@sohaibhasan Merging this in, tagging you for fidelity and in case you have back comments.

PR #449, opened 2017-11-17
Using cursor.

sheyd:
#### What's this PR do?
* Updates ingestion from Rogue to use the cursor parameter.
* Updates SQL schema to use `root` instead `quasaradm` for ingestion.
One file below, 2 actually.
Checkout and run on local vagrant.
- Current ingestion brings Rogue to its knees.
https://www.pivotaltracker.com/story/show/152944442

### DIFF
--- a/data/sql/migrations/000_schema_quasar.sql
+++ b/data/sql/migrations/000_schema_quasar.sql
@@ -54,7 +54,7 @@ CREATE TABLE `campaign_activity` (
 
 DELIMITER ;;
 /*!50003 SET SESSION SQL_MODE="NO_ENGINE_SUBSTITUTION" */;;
-/*!50003 CREATE */ /*!50017 DEFINER=`quasaradm`@`%` */ /*!50003 TRIGGER `log` AFTER INSERT ON `campaign_activity` FOR EACH ROW BEGIN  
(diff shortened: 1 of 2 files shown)


data/sql/migrations/000_schema_quasar.sql
```diff
@@ -54,7 +54,7 @@ CREATE TABLE `campaign_activity` (
 
 DELIMITER ;;
 /*!50003 SET SESSION SQL_MODE="NO_ENGINE_SUBSTITUTION" */;;
-/*!50003 CREATE */ /*!50017 DEFINER=`quasaradm`@`%` */ /*!50003 TRIGGER `log` AFTER INSERT ON `campaign_activity` FOR EACH ROW BEGIN  
+/*!50003 CREATE */ /*!50017 DEFINER=`root`@`%` */ /*!50003 TRIGGER `log` AFTER INSERT ON `campaign_activity` FOR EACH ROW BEGIN
```
DFurnes: Why is this changing?

sheyd: For local dev, the `quasaradm` user doesn't exist in the Vagrant MySQL instance, which is what these files are used for. So just resetting to root :)

DFurnes: Got it! 👍

PR #451, opened 2017-11-27
Adding page tracking for large backfills.

sheyd:
#### What's this PR do?
Updates campaign activity ingestion so if hourly backfills are over 120 hours (5 days) _or_ we do a full backfill, page tracking for the backfill comes into play. This allows us to resume large backfills from paginated progress, and not from the beginning and get stuck again.

This is admittedly a bit of an arbitrary hack, but will allow us to resume backfills from large numbers of campaign activity.

One file below.
Tested locally with both > 120 and < 120 backfills, and DB updates work.

Large backfill for SMS activity is stuck at page 82464, and we want to progress beyond that.
https://www.pivotaltracker.com/story/show/152944442

### DIFF
--- a/quasar/campaign_activity.py
+++ b/quasar/campaign_activity.py
@@ -35,11 +35,11 @@ def _backfill(hours=None):
                               'filter[updated_at]': start_time,
                               'pagination': 'cursor'})
 
-    if hours is not None:
+    if int(hours) > 120 or hours is None:
+        current_page = _get_start_page(db)
+    else:
         print("Current backfill hours are %s." % hours)
         current_page = 1
-    else:
-        current_page = _get_start_page(db)
 
     page = scraper.getJson('', params={'page': current_page})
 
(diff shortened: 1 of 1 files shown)


quasar/campaign_activity.py
```diff
@@ -35,11 +35,11 @@ def _backfill(hours=None):
                               'filter[updated_at]': start_time,
                               'pagination': 'cursor'})
 
-    if hours is not None:
+    if int(hours) > 120 or hours is None:
```
sohaibhasan: Is using int the best thing to do here? And is this the best place to do it? I can't see where hours is initially set, but would it be better to do any rounding and coercion there?

sheyd: `Int` is used since Python 3 doesn't auto cast the variable (which is presented as a `string` at this point) when evaluating the conditional. 

Hours is set (or not), from function calls: https://github.com/DoSomething/quasar/pull/451/files#diff-0b03649b386ffaed2898b4f486b7ec1eR15 or https://github.com/DoSomething/quasar/pull/451/files#diff-0b03649b386ffaed2898b4f486b7ec1eL23

And also, not trying to refactor too much here, since we're going to be moving to the push to queue model next quarter.

sohaibhasan: Fair enough

PR #468, opened 2018-01-09
Adding x-max-priority to queue declaration.

sheyd:
#### What's this PR do?
Blink added priority functionality to queues. As the default is `2`, it's necessary to define the priority variable for the Quasar queue, as the default of `None` returns an error.
One file below.
Checkout and run using `runscope_cleanup` test script.
First PR of 2018!
https://www.pivotaltracker.com/story/show/154184866

### DIFF
--- a/quasar/queue.py
+++ b/quasar/queue.py
@@ -31,7 +31,8 @@ class QuasarQueue:
         self.connection = pika.BlockingConnection(self.params)
         self.channel = self.connection.channel()
         self.channel.basic_qos(prefetch_count=config.QUEUE_PREFETCH_COUNT)
-        self.channel.queue_declare(amqp_queue, durable=True)
+        self.channel.queue_declare(amqp_queue, durable=True, 
+                                   arguments={'x-max-priority' : 2})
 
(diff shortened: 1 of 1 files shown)


blisteringherb:
There's a Stack Overflow article saying that these priorities need to be declared when the queue is created. Does that apply to this? https://stackoverflow.com/questions/10745084/rabbitmq-and-message-priority Otherwise, 👍

rapala61:
Interesting. I think this is what Sergii was referring to in the PR description:
>Assert a queue into existence. This operation is idempotent given identical arguments; however, it will bork the channel if the queue already exists but has different properties (values supplied in the arguments field may or may not count for borking purposes; check the borker’s, I mean broker’s, documentation). - [Link](https://github.com/DoSomething/blink/pull/181).

sheyd:
Yeah, Rafa manually recreated the queue today with that priority, so the consumer isn't creating the queue, just attaching to it already.

sheyd:
Ahhh, interesting note, Rafa, thanks for pointing that out. Confirms implementation in this PR fixes things, and matches spec :)

PR #473, opened 2018-01-22
Rogue test consume route

sheyd:
#### What's this PR do?
- Updates base Queue class to have RabbitMQ channel setting `confirm_delivery` set, so published messages are guaranteed and acknowledged delivery. Increasingly important as we use publish methods in more places. Noticed during testing message loss at times.
- Adds testing class for publish random test and Rogue test messages using the `[data][meta][message_source]` field. This is what the c.io consumer code will use to filter messages from Rogue once Rogue starts publishing to Quasar queue via Blink.

https://github.com/DoSomething/quasar/compare/rogue-test-consume-route?expand=1#diff-746654cf6158b50e97892d2b43b93f26.

Checkout the PR, and with proper config file values, run:
- `rogue_generic_pub`
- `rogue_test_pub`
- `rogue_test_route`
- `rogue_test_consume`

Checking that 100 messages of each type generic and Rogue get sent to `quasar.test` queue (200 total). Check after `rogue_test_route` that 100 messages of type Rogue are in `quasar.rogue` queue. Run `rogue_test_consume` and confirm consuming of all `quasar.rogue` queue messages.

This testing script was used to test filtering code that will be used in Production shortly or filtering out Rogue messages from C.IO inbound messages.

https://www.pivotaltracker.com/story/show/154369417


quasar/rogue_queue_test.py
```diff
@@ -0,0 +1,73 @@
+import json
+import time
+
+from .config import config
+from .queue import QuasarQueue
+
+
+class RogueQueue(QuasarQueue):
```
DFurnes: Should this class be in here, or would it make more sense in a separate `rogue_queue.py`?

quasar/rogue_queue_test.py
```diff
@@ -0,0 +1,73 @@
...
+        super(RogueQueue, self).__init__(config.AMQP_URI, config.ROGUE_QUEUE,
+                                         config.QUASAR_EXCHANGE)
+
+    def process_message(self, message_data):
+        message = json.loads(message_data)
+        print(''.join(("Timestamp {} for Rogue message consumed."
+                       "")).format(message['data']['data']['timestamp']))
+
+
+rogue_queue = RogueQueue()
```
DFurnes: It might make sense to put these in each of the "script runner" methods at the bottom, so that way it's only instantiated if it's needed for the particular script that's being run (and is closer to where it's actually used):

```py
def test_consume():
    queue = TestQueue()
    queue.start_consume()

def rogue_consume():
    queue = RogueQueue()
    queue.start_consume()

```

PR #474, opened 2018-01-23
Rogue Message Filtering

sheyd:
#### What's this PR do?
If a message with `[data][meta][message_source]` key exists, it signals there's a message that's published from Rogue via Blink into the Quasar primary queue. This PR filters the message into a dedicated Rogue queue.

This is the important file: https://github.com/DoSomething/quasar/compare/rogue-message-filter?expand=1#diff-d058ae5dd8b86ca1e1eaa7e6d1d138f8

Everything else is from https://github.com/DoSomething/quasar/pull/473 and what's in the `master` branch.

Will test manually in prod, since currently don't have staging environment. Testing happened in controlled environment with test data in https://github.com/DoSomething/quasar/pull/473.

We need to get this into production this sprint so Rogue isn't blocked from pushing messages into Quasar queue for voter reg data.

https://www.pivotaltracker.com/story/show/154369417


quasar/cio_queue_process.py
```diff
@@ -10,21 +10,36 @@
 db = Database()
 
 
+class RogueQueue(QuasarQueue):
```
DFurnes: Should this import the same `RogueQueue` used in #473, rather than defining a different one per file?

PR #489, opened 2018-02-08
Rogue consumer

sheyd:
#### What's this PR do?
What doesn't this PR do is a better question. 🎃 

- Rogue event consumer so all data pushed from Rogue can be used to update the `campaign_activity` table.
- Handles new event types and fields, like post details and deletion of signups and posts.
- Bumps Quasar version to 0.2.0, since this is our first dedicated non-CIO event consumer for one of our internal systems.
- Puts all consumer code into a `quasar_queue` library class, and end-user ETL runtimes handle creating and initializing actual consumers.
- Updates dev SQL schema based on latest Quasar MySQL DB tables and fields, as well as deprecating older ones we no longer use.
[Here]https://github.com/DoSomething/quasar/compare/rogue-consumer?expand=1#diff-68da9c0535d3768969007f65eaf1ffd9l) is probably the best place.

Multiple rounds of testing have occurred on staging with @katiecrane. 
This will make everyone's lives 3147% better. 
https://www.pivotaltracker.com/story/show/154740170


data/sql/migrations/quasar-schema-dump.sql
```diff
@@ -0,0 +1,733 @@
+-- MySQL dump 10.13  Distrib 5.7.21, for osx10.13 (x86_64)
```
DFurnes: It'd be nice to be able to track changes to this schema as actual migrations eventually. It looks like `alembic` might be a pretty good option, and [can automatically generate them from your actual schema](https://www.compose.com/articles/schema-migrations-with-alembic-python-and-postgresql/)!

quasar/quasar_queue.py
```diff
@@ -0,0 +1,245 @@
...
+        self.db.query_str(''.join(("DELETE FROM ",
+                                   self.campaign_activity_table,
+                                   " WHERE status = %s AND "
+                                   "signup_id = %s AND "
+                                   "signup_updated_at = %s")),
+                          ('signup_deleted', signup_id, deleted_at))
+        print("Signup {} deleted and archived.".format(signup_id))
+
+    def _add_post(self, post_data):
+        self.db.query_str(''.join(("REPLACE INTO ",
```
DFurnes: These huge `query_str` calls make me nervous. It'd be nice to start experimenting with [sqlalchemy](http://www.sqlalchemy.org) or [pypika](https://github.com/kayak/pypika) to build these for us (even just in a few places). Otherwise it seems so easy for columns & values to get mixed up.

sheyd: Following up no IRL convo with @DFurnes about this yesterday. Added to the MySQL -> PostgreSQL migration task list to see where we can start using SQL Alchemy in a low lift way.

data/sql/migrations/quasar-schema-dump.sql
sheyd: Also talked to @DFurnes about this IRL, and would love to implement this. Adding to MySQL -> PostgreSQL list!

PR #494, opened 2018-02-16
Adding Postgres to Vagrant box install

blisteringherb:
#### What's this PR do?
This PR adds Postgres 10 to the Vagrant dev box

Run `vagrant destroy` locally, and then `vagrant up` with these changes to see if Postgres is installed and running

https://www.pivotaltracker.com/story/show/155147874


sheyd:
Had to add 1 fix for MySQL import `provision.sh` line to only import `*.sql` files, not any others that might appear in that directory.

PR #495, opened 2018-02-23
Updating post details parsing code.

sheyd:
#### What's this PR do?
- Fixes logic around checking whether a post has details for not to ingest from double-negative, to much more straight-forward checking if the field exists or has a blank value.
- Fixes ghost bug that we hadn't encountered yet that didn't properly assign a variable value.
One file below.
- Oh you know... tested in Prod out of desperation.

### DIFF
--- a/quasar/quasar_queue.py
+++ b/quasar/quasar_queue.py
@@ -265,8 +265,8 @@ class RogueQueue(QuasarQueue):
         elif data['meta']['type'] == 'post':
             if not pydash.get(data, 'deleted_at'):
                 self._add_post(data)
-                if ((pydash.get(data, 'details') is None) or
-                       (pydash.get(data, 'details') == '')):
+                if (pydash.get(data, 'details') is None or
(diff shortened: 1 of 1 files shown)


quasar/quasar_queue.py
```diff
@@ -173,6 +173,8 @@ def _add_post_details(self, post_id, post_details):
         # TODO: Remove type check if Rogue sends this as JSON/dict.
         if type(post_details) is str:
             details = json.loads(post_details)
+        else:
+            details = post_details
```
DFurnes: It might be clearer to just pass in `details` as the argument rather than redefining as a local var:

```python
def _add_post_details(self, post_id, details):
    # TODO: Remove type check if Rogue sends this as JSON/dict.
    if type(details) is str:
        details = json.loads(post_details)

    if pydash.get(details, 'source_details'):
        # and so on...
```

sheyd: Merging to get the fix out, but will add this to different update branch.

PR #501, opened 2018-03-13
Rogue Postgres Consumer

sheyd:
#### What's this PR do?
* Creates Rogue consumer for events to get loaded into Postgres based data warehouse.
* Simplifies `pydash.get` check logic from `if not` to `if` as discussed in another PR with @DFurnes. Specifically see https://github.com/DoSomething/quasar/blob/rogue-pg-consumer/quasar/quasar_queue.py#L263-L265 vs https://github.com/DoSomething/quasar/blob/rogue-pg-consumer/quasar/quasar_queue.py#L430-L432.
* Bumps version to `0.4.0`

https://github.com/DoSomething/quasar/blob/rogue-pg-consumer/quasar/quasar_queue.py#L283

I was able to successfully test consuming Blink staging `quasar.rogue-pg` messages to my local DB successfully, which had ~11000 events, including a signup delete and uploaded post_details.

We're migrating from MySQL to Postgres, and this should handle the Rogue ingestion point.

https://www.pivotaltracker.com/story/show/155919704

### DIFF
--- a/quasar/quasar_queue.py
+++ b/quasar/quasar_queue.py
@@ -361,7 +361,7 @@ class RoguePostgresQueue(QuasarQueue):
                                    "(id, created_at, updated_at, "
                                    "status, deleted_at) VALUES "
                                    "(%s,%s,%s,%s,%s)")),
-                          (post_id, created_at[0], deleted_at, 
+                          (post_id, created_at[0], deleted_at,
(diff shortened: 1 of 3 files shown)


sheyd:
@DFurnes Tagging you on this since you have more code and logic context to this than other team ⚡️ members.

PR #510, opened 2018-03-15
Remove old commands

sheyd:
#### What's this PR do?
* Removes a metric 💩 🚚 of old libraries and commands from the era of MoCo as well as the last Josh Cusano code. ☠️
* Updates documentation to be macOS focused only and use `virtualenv` instead of `pyenv`.
* Fixes the red double compile errors we saw on macOS.
* Bumps version to `0.5.0`
https://github.com/DoSomething/quasar/compare/remove-old-commands?expand=1#diff-04c6e90faac2675aa89e2176d2eec7d8
Checked all code, and none of it is necessary locally.
We've moved to a push centric model, and with @blisteringherb moving to the ⚡️ team, I'd like to have a fresh a start as possible. Also, getting a consistent dev environment setup is key.


README.md
```diff
@@ -74,17 +84,23 @@ cd $QUASAR_PROJECT_DIR
...
+
 ## Usage
 
 ```
 cd $QUASAR_PROJECT_DIR
 pip install -e .
+make build
 ```
 
 See `setup.py` for list of entry-points. E.g.
```
blisteringherb: I know this isn't a tutorial for novices necessarily, but how would you feel about adding a bit more explanation about what entry points might be?

sheyd: @blisteringherb added some more clarification to the docs in the latest commit.

PR #540, opened 2018-05-07
Adding indexes to Puck Mat views.

sheyd:


sohaibhasan:
@sheyd I think this PR should be redundant now, yeah?

PR #551, opened 2018-06-05
Bertly refresh

sheyd:
#### What's this PR do?
* Refreshes sms_click data from Bertly by kicking of DMS full refresh.
https://github.com/DoSomething/quasar/compare/bertly-refresh?expand=1#diff-1830236ee3d3073b76b0bee40244fc93
Tested against DMS instance live multiple times.
* This is a workaround until AWS can fix perpetual replication.
* We also still likely need a Mat View to make sure derived tables from this are not blocked by DB refreshes, which drops the whole source table.
https://www.pivotaltracker.com/story/show/157194637


sheyd:
I'm ignoring the somewhat ridiculous linting on this one. Once PR is approved, will merge.

PR #574, opened 2018-07-19
adding caption to post and campaign activity derived tables

sohaibhasan:
#### What's this PR do?
* Adds rogue.posts.caption to public.posts and public.latest_post and public.campaign_activity

* campaign_activity.sql


jamjensen:
Is it still labeled as `caption` -- I thought you said it was renamed as `text`?

sohaibhasan:
@jamjensen Other way around. Rogue calls it `text` but we call it `caption`. I can rename it to `text` if that is preferred

PR #577, opened 2018-07-30
Phoenix campaigns

sheyd:
#### What's this PR do?
* Adds data pipeline to query for all campaigns id's from `rogue.signups` table, then query the Phoenix API for full campaign data. This is to replace our current Ashes based campaign_info table solution.
* Updates Scraper class to remove XML parsing, since that was legacy from MoCo days.
* Adds generic DS_OAUTH_Scraper class so that we can use Northstar auth to multiple services and key refreshes are automatically handled.
* Bumps version to `1.4.0`.
* Also now with 💯 more logging than we've ever had before instead of `print` statements!

https://github.com/DoSomething/quasar/compare/phoenix-campaigns?expand=1#diff-9c0c7201c8000c8539f08ed404e0c4b4
Tested in Quasar QA db with no issues.
This is a medium-term solution until Phoenix is completely migrated to and there is a scrapable paginated API for all campaigns.
https://www.pivotaltracker.com/story/show/159239490

### DIFF
--- a/quasar/campaign_info.py
+++ b/quasar/campaign_info.py
@@ -1,0 +1,6 @@
+import os
+from .database import Database
+from .ds_oauth_scraper import DSOAuthScraper
+
+scraper = 
+
--- a/quasar/campaign_info.py
+++ b/quasar/campaign_info.py
@@ -1,6 +1,8 @@
 import json
 import logging
 import os
+import sys
+
 from .database import Database
 from .ds_oauth_scraper import DSOAuthScraper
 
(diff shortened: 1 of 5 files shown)


quasar/campaign_info.py
```diff
@@ -0,0 +1,55 @@
...
+                db.query_str(''.join(("INSERT INTO phoenix.campaigns_json "
+                                      "(campaign_data) VALUES (%s)")),
+                             (json.dumps(info),))
+                logging.info('Campaign {} added to db.'.format((id[0])))
+
+            else:
+                logging.error(''.join(('{} status code returned for campaign '
+                                      '{}.'.format(result.status_code,
+                                                   id[0]))))
+    except:
```
DFurnes: Would it make sense to move this try/catch block closer to the `scraper.get` call so that it can continue to try to fetch other campaigns if it hits an error on one of them?

quasar/ds_oauth_scraper.py
```diff
@@ -0,0 +1,35 @@
...
+    def authenticated(func):
+        def _authenticated(self, *args, **kwargs):
+            response = func(self, *args, **kwargs)
+            if response.status_code == 401:
+                self.auth_headers = self.fetch_auth_headers()
+                response = func(self, *args, **kwargs)
+            return response
+        return _authenticated
+
+    @authenticated
```
DFurnes: This is really clean! 🔏

quasar/campaign_info.py
sheyd: Talked through with Dave IRL, and decided to leave current exception catch-all as is for MVP implementation.

PR #595, opened 2018-08-27
Updating references to relationized gambit_conversations tables to have _flattened appended to table names.

sheyd:
#### What's this PR do?
* Updated table references in `gambit_conversations` that we've extrapolated out to relational format to have `_flattened` appended to them, since DMS table renaming has caused us issues.

https://github.com/DoSomething/quasar/compare/gambit-convo-fix?expand=1#diff-3093009998c9784257738b2d8478c353
Ran in staging.

https://www.pivotaltracker.com/story/show/159699166

### DIFF
--- a/data/sql/derived-tables/gambit_conversations.sql
+++ b/data/sql/derived-tables/gambit_conversations.sql
@@ -1,12 +1,17 @@
 DROP MATERIALIZED VIEW IF EXISTS gambit_conversations.conversations_flattened;
 CREATE MATERIALIZED VIEW gambit_conversations.conversations_flattened AS
 (SELECT
-   (records->>'campaignId')::varchar as campaign_id, to_timestamp((((records->'createdAt'->>'$date')::BIGINT) / 1000)::double precision) as created_at,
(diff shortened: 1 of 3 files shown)


data/sql/derived-tables/gambit_conversations.sql
```diff
@@ -1,5 +1,5 @@
-DROP MATERIALIZED VIEW IF EXISTS gambit_conversations.conversations;
-CREATE MATERIALIZED VIEW gambit_conversations.conversations AS
+DROP MATERIALIZED VIEW IF EXISTS gambit_conversations.conversations_flattened;
+CREATE MATERIALIZED VIEW gambit_conversations.conversations_flattened AS
 (SELECT
    (records->>'campaignId')::varchar as campaign_id, to_timestamp((((records->'createdAt'->>'$date')::BIGINT) / 1000)::double precision) as created_at,
```
sohaibhasan: I think if you use `->` instead of `->>` it automatically casts as text

sohaibhasan: Can we also make this a little prettier where each column is in another row

PR #624, opened 2018-09-11
Adding file to run daily Gambit DMS import

blisteringherb:
#### What's this PR do?
The Gambit streaming in the DMS wasn't working so we have to run a daily import.


quasar/gambit_import.py
```diff
@@ -0,0 +1,65 @@
...
+    while not (status['status'] == 'stopped' and
+               status['reason'] == 'Stop Reason FULL_LOAD_ONLY_FINISHED' and
+               status['progress'] == 100):
+        # SLeep between checks for 60 seconds, then try again.
+        log("Gambit DMS refresh still not done, waiting.")
+        time.sleep(60)
+        status = check_refresh_status()
+
+    db = Database()
+
```
sheyd: Do we need the Database class and refresh of mat views if we're going to trigger a downstream Jenkins job for mat view refreshes?

PR #627, opened 2018-09-13
Gambit recreate test

sheyd:
#### What's this PR do?
* Updates the SQL for Gambit Messages by removing comment lines and updating permissions grants.
* Adds dependency run of Gambit Conversations create before Gambit Messages, since Gambit Messages derived tables rely on Gambit Conversations derived tables.
* Added logging for individual activities.

One outstanding question @sohaibhasan @blisteringherb, I think it's a good idea to include the `*_json` creation statements here (we can include as a separate file). We've kind of treated them as stored procedures at this point, and copy-pasted them from tickets each time we've created, but I'd love to capture them as code. That being said, the creation takes a while, so should we capture them in a separate function like `create_gambit_base_table`, or add it as a line to the re-create?

https://github.com/DoSomething/quasar/compare/gambit-recreate-test?expand=1#diff-ee3906a65114da5425775eb8dc25047a
It's been run in staging and prod for final testing.

https://www.pivotaltracker.com/story/show/160421340


data/sql/derived-tables/gambit_messages.sql
```diff
@@ -126,7 +120,8 @@ LEFT JOIN
 	AND u.mobile <> ''
 WHERE
 	g.direction <> 'inbound'
-	AND g.user_id IS NULL )
-);
+	AND g.user_id IS NULL));
 
 CREATE INDEX outbound_messages_i ON public.gambit_messages_outbound (message_id, created_at, user_id, conversation_id);
+GRANT SELECT ON public.gambit_messages_outbound to looker;
+GRANT SELECT ON public.gambit_messages_outbound to dsanalyst;
```
blisteringherb: Not sure if you want to add a newline here.

blisteringherb:
I'm inclined to agree in terms of having the recreate json table code in the repo. I think it'll help document the whole process better also. I see you've removed some comments, but it would be nice if we could figure out a way to use comments to automatically document the code Doxygen used to do.

PR #637, opened 2018-09-16
Fixing missing , in setup.py.

sheyd:
#### What's this PR do?
Build is breaking due to missing ',' in setup.py. Fixes this.
https://github.com/DoSomething/quasar/compare/setup-syntax-fix?expand=1#diff-2eeaed663bd0d25b7e608891384b7298


blisteringherb:
I can actually take all this code out if you like. I put it in so we could debug, but since it looks like the Gambit import is working fine, we don't need it at the moment.

sheyd:
Sounds good!

PR #648, opened 2018-10-23
Update sql logic in bertly and mel

mjain-1:
#### What's this PR do?
It updates the bertly logic to capture Android and iOS previews in the interaction type field and appends whether the action is a click or preview in the mel logic. 
Check bertly.sql followed by mel.sql.
This is part of the effort to identify clicks vs. autogenerated previews vs. user-loaded previews. Given that there is no way to differentiate between the latter two using the user agent string, the decision was made to eventually ignore previews. This PR is the first step in that direction by labeling interactions as clicks/previews in bertly and then in mel.
https://www.pivotaltracker.com/story/show/161419368

### DIFF
--- a/data/sql/derived-tables/bertly.sql
+++ b/data/sql/derived-tables/bertly.sql
@@ -21,8 +21,8 @@ CREATE MATERIALIZED VIEW public.bertly_clicks AS (
 			END) AS SOURCE,
 		CASE 
 			WHEN c.user_agent IS NULL THEN 'uncertain'
-			WHEN c.user_agent ILIKE '%%facebot%%' 
-				OR c.user_agent ILIKE '%%twitterbot%%' THEN 'preview'
+			WHEN c.user_agent ILIKE '%%facebot% twitterbot%'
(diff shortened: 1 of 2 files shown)


data/sql/derived-tables/bertly.sql
```diff
@@ -21,8 +21,8 @@ CREATE MATERIALIZED VIEW public.bertly_clicks AS (
 			END) AS SOURCE,
 		CASE 
 			WHEN c.user_agent IS NULL THEN 'uncertain'
-			WHEN c.user_agent ILIKE '%%facebot%%' 
-				OR c.user_agent ILIKE '%%twitterbot%%' THEN 'preview'
+			WHEN c.user_agent ILIKE '%%facebot% twitterbot%'
```
sohaibhasan: We have to surround these greps with double `%` signs because of some python mumbo jumbo. @sheyd can add color

mjain-1: 👍 got it, making that now!

PR #661, opened 2018-11-19
Mel query public tables

mjain-1:
#### What's this PR do?
It generates the member event log using the public.posts and public.signups tables rather than rogue, allowing for the public tables to be used as a cleaned source for both MEL and campaign activity. Given that the public.posts table now has the correct registered date for RTV/TV records, we no longer need to pull them from campaign activity (allowing us to remove the final union).   

There are three changes: 1) pointing to the public.signups table, 2) pointing the public.posts table, and 3) removing the union to campaign activity. 

The impact to MAMs will be minimal: ~1 MAM lost in 4-5 months, an increase in June (~12 MAMs) from the now included TV records.

### DIFF
--- a/data/sql/derived-tables/mel.sql
+++ b/data/sql/derived-tables/mel.sql
@@ -17,71 +17,33 @@ FROM (
         '1' AS action_id, 
         s."source" AS "source",
         s.id::varchar AS action_serial_id,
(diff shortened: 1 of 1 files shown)


sohaibhasan:
Love all the deprecated code!

PR #673, opened 2018-12-07
Updating campaign_info mat view refresh from Rogue.

sheyd:
#### What's this PR do?
- Removes no longer used Phoenix based campaign_info file.
- Updates `campaign_info`, `campaign_info_all`, and `campaign_info_international` SQL to create mat views.
- Updates `campaign_info` refresh with appropriate mat views.
SQL manually created in Prod.
We're not longer relying on Phoenix for campaign info data. It has been supplanted by Rogue.
https://www.pivotaltracker.com/story/show/162313317


data/sql/derived-tables/campaign_info.sql
```diff
@@ -40,30 +41,30 @@ CREATE TEMPORARY TABLE IF NOT EXISTS campaign_info_all AS (
     GROUP BY 1,2,3,4,5,6,7,8,9,10,11 
     ORDER BY c.field_campaigns_target_id, fdfrd.field_run_date_value);
     
-DROP TABLE IF EXISTS campaign_info;
-CREATE TABLE IF NOT EXISTS campaign_info AS (
+DROP MATERIALIZED VIEW IF EXISTS public.campaign_info;
+CREATE MATERIALIZED VIEW public.campaign_info AS (
```
sohaibhasan: Should we turn this into `CREATE MATERIALIZED VIEW IF NOT EXISTS`? Seems to be serving us well in other cases

sheyd: I'm happy to add it in, but seemed redundant given the drop statement the line before. However, does no harm so adding it back in.

PR #675, opened 2018-12-07
Almost works. Needs to cascade delete.

sheyd:
#### What's this PR do?
- Adds cascade to Mat View recreate for `campaign_info` tables.


sheyd:
Hmm, good thought. I'll create a ticket for it.

PR #676, opened 2018-12-10
New reportback table

mjain-1:
#### What's this PR do?
This PR refactors campaign_activity in a couple of ways outlined in detail [here](https://docs.google.com/spreadsheets/d/1rK3GqV2VoLYW8m8RKVP9Lvx7EKthzHne7ZA9aqIaOQE/edit#gid=0)) and summarized below:

- It moves fields currently only contained in CA to the signups, latest_post, and posts tables, as well as adds fields to the posts table identifying accepted and rb posts.
- It removes campaign-specific rules per management's decision on rb until post metadata is available (and that logic will need to be updated).
- It replaces the `campaign_activity` table with a `reportbacks` table that contains only accepted rbs.

reportbacks.sql

I've confirmed that September and October rb counts are overall the same. There is one change related to removing the `post_attribution_date` field: In the case of a user submitting multiple posts corresponding to multiple drives for voter reg, previously all counts were attributed to the date of the first post, while now counts are associated to the date/month of when the post is created. This only affects users who submitted drive counts across multiple months. 

Further testing will be done to confirm impact once tables created in prod, as well as via Looker.

https://www.pivotaltracker.com/story/show/162421483

#### Questions:

- Should this logic still account for current campaign-specific rules implemented in `campaign_activity`?
- Thoughts on having separate `is_accepted` and `is_reportback` fields?
- Thoughts on indexes? I took a stab at creating some that might be helpful but happy to keep things as is if we don't want to make too many changes here.

### DIFF
--- a/data/sql/derived-tables/mel.sql
+++ b/data/sql/derived-tables/mel.sql
@@ -26,10 +26,6 @@ FROM (
     WHERE s."source" IS DISTINCT FROM 'importer-client'
 	AND s."source" IS DISTINCT FROM 'rock-the-vote'
 	AND s."source" IS DISTINCT FROM 'turbovote'
-	AND s.id NOT IN (SELECT c.signup_id AS id
-    				FROM campaign_activity c 
-    				WHERE c.signup_source = 'importer-client' 
-    				AND c.signup_created_at > c.post_created_at)
     UNION ALL
     SELECT -- CAMPAIGN POSTS WITH CHANNEL
         DISTINCT p.northstar_id AS northstar_id,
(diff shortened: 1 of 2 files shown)


data/sql/derived-tables/mel.sql
```diff
@@ -26,10 +26,6 @@ FROM (
     WHERE s."source" IS DISTINCT FROM 'importer-client'
 	AND s."source" IS DISTINCT FROM 'rock-the-vote'
 	AND s."source" IS DISTINCT FROM 'turbovote'
-	AND s.id NOT IN (SELECT c.signup_id AS id
```
mjain-1: In addition to removing the dependency on CA, we don't need this logic because it's already being captured by excluding signups created by importer-client. I confirmed that this does not impact overall count.

PR #679, opened 2018-12-13
update puck etl json to account for old and new event names

sohaibhasan:
#### What's this PR do?
* Updates the calculation of puck events event name to account for new event names
* Creates a new field for the new name so transition to this field cannot be done asynchronously to rocket changes
* create_puck_etl_json.sql
* Run SELECT that build phoenix events
* Does every event have a name?
* Rocket is rolling out new names for puck events, but don't currently have bandwidth to backfill historical data with these values. We also cannot update on our end since the entire database is synced nightly. 
* To account for this we are given a lookup table translating old event names to new event names
* LOGIC:
  1) The logic is to join this lookup table twice to the events table, first by the old name, then by the new name. 
  2) If joining by the old name gives a result (meaning the event's name reflects the old name, the event's name is the old name
  3) If this condition is not met, then the event must have the new name, so joining by new name must have succeeded, and so, we choose the old name column resulting from this join. 
* Thus, we have the existing event names for all events, obviating changes to any views in Looker or existing queries
* https://www.pivotaltracker.com/story/show/162649091
* https://www.pivotaltracker.com/story/show/162068080
#### Questions:
* Does this logic make sense?

### DIFF
--- a/data/sql/derived-tables/create_puck_etl_json.sql
+++ b/data/sql/derived-tables/create_puck_etl_json.sql
@@ -48,7 +48,12 @@ CREATE MATERIALIZED VIEW public.phoenix_events AS (
 		e.records #>> '{meta,id}' AS puck_id,
 		to_timestamp((e.records #>> '{meta,timestamp}')::bigint/1000) AS event_datetime,
 		(e.records #>> '{meta,timestamp}')::bigint AS ts,
-		e.records #>> '{event,name}' AS event_name,
+		CASE 
+			WHEN enames_old.old_name IS NOT NULL 
+			THEN enames_old.old_name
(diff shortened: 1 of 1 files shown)


sheyd:
Oh dear gods, just let me build it as a mat view or snapshot table with a different name, then you can query/index against it to your hearts content.

sheyd:
@sohaibhasan generated `public.phoenix_events_newname` in Prod with indexes to test against. 😈

sohaibhasan:
@sheyd Counts look good on my end, though the real test of course isn't until these events actually start showing up. I'm good to deploy and keep a close eye over the next couple of days.

PR #711, opened 2019-02-05
Fivetran gambit

blisteringherb:
#### What's this PR do?
This PR adds updated sql files and updated tasks to pull the data from the Fivetran tables into materialized views for consumption by Looker. I've duplicated all of the sql from the sql files that generate the data mat views for Gambit so that we can test this independently without interfering with the current setup.
https://www.pivotaltracker.com/n/projects/2076969/stories/163565949
#### Questions:
Please take a look and make sure the naming looks ok for now, keeping in mind that we'll likely have another clean up task which may result in renaming some or all of the tables once we feel confident this code is generating the correct info.


blisteringherb:
@sheyd I'd like to remove the gambit_conversations_ft.sql file since it doesn't really do anything anymore. We don't need to generate a conversations_flattened table since the updated FT code can just use the new ft_gambit_conversations_api.conversations table. That's one less derived table. The only thing I can find in the codebase that uses that table is Gambit messages and the ETL monitoring. Any objections?

sheyd:
@blisteringherb None whatsoever. :)

PR #727, opened 2019-03-05
Puck rogue cleanup

sheyd:
#### What's this PR do?
* Updates `campaign_info` tables to use Fivetran Rogue propagation instead of old DMS job.
* Updates `phoenix_events` and `phoenix_sessions` to refresh concurrently since they're long-running, and non-current refreshes were blocking Looker view rebuilds.
https://github.com/DoSomething/quasar/compare/puck-rogue-cleanup?expand=1#diff-e265a6c36f85b94b24384760001a8f3e
* Queries are already tested, this is just reference updating. But will manually test after hours today.


sohaibhasan:
Good catch! Totally forgot about this table

PR #746, opened 2019-04-03
Cio staging import

sheyd:
#### What's this PR do?
* Updates C.io consumer to write to `_staging` tables.
* Create C.io importer to from staging tables to primary queried tables.
* Updates `requirements.txt` to use Python 3.7 runtime since that's what we're running everywhere now.
* Some renaming of `cio_import` to `cio_consume`, and new importing code to `cio_import`.
https://github.com/DoSomething/quasar/compare/cio-staging-import?expand=1#diff-1ab043edaf34692e48c60af8bd2b6dab
In QA!
https://www.pivotaltracker.com/story/show/164869907


blisteringherb:
In the past and currently in some applications we refer to environments as "staging", so I'm wondering if we should consider changing these table names to "scratch" or "landing zone" tables so that we avoid confusing when we have `_staging` tables in production databases.

sheyd:
This is a great thought! I'll alter the tables in the backend to be `_scratch`, and update the code!

PR #778, opened 2019-06-07
Add post tags to posts

sohaibhasan:
#### What's this PR do?
* pulls in post tag information into public.posts and public.reportbacks
* campaign_activity.sql
* Run public.post and public.reportback SELECT statements
* Campaigns requested that these be added as Looker dimensions, followup Looker PR to follow
* Trello card incoming
#### Questions:
* Could use input on how table referencing should work. I see we're doing some aliasing, but not sure how that all works


sohaibhasan:
@sheyd It looks like campaign activity recreate happened in 55 seconds in the last run?

PR #783, opened 2019-06-13
Adding first snowplow_base_event table sql.

sheyd:
#### What's this PR do?
* Creates Snowplow raw event tables necessary for combined Puck + Snowplow `phoenix_events` table.
* Uses stage_table/truncate/insert method for table refreshes instead of materialized views as explained below.

* Tested successfully from my laptop with QA warehouse values. 

Traditionally, we've used `refresh materialized view` for propagating new/refresh data into our raw and derived data layers. As our data volume and table sizes have grown, we've started to hit the limitations of this [simple but ultimately non-scaleable method](https://www.citusdata.com/blog/2018/10/31/materialized-views-vs-rollup-tables/). 

We've discussed using an approach that utilizes staging tables that re-run the queries, truncates existing utilized tables, and inserts all the newly calculated data. In essence:
`create table_stage` -> `truncate table` -> `insert * from table_stage to table`

While this approach has some obvious limitations (re-calculating the new full data set each day as data is added), it provides us with more atomic controls, and reduces table contention since while `truncate` requires an `ACCESS EXCLUSIVE` table lock, this operation is much shorter than the extended table locks that `refresh mat view` requires. We also get the benefit of not having as much `VACUUM` maintenance since truncate releases all table disk blocks immediately.

For the long run, it likely makes sense to use date based roll-ups as noted in the article above, but this will require significantly more thinking on our part for how we calculate derived tables, and is more work than we have time for now, and this allows us to have a better approach in the interim.

Some points to note is that this first attempt uses a very copy/paste replication of SQL code approach that isn't variablized. When we roll this out to other derived tables in Q3, we should consider a more consistent, reproducible series of functions and variable names. In the meantime, we have to make sure we keep table modifications accounted for in both `snowplow_events.sql` and `snowplow_events_refresh.sql` files. 

https://www.pivotaltracker.com/story/show/166161952


sheyd:
One other thing I've added is `pipenv` so we can migrate away from using `requirements.txt`.

PR #784, opened 2019-06-26
Adding first draft of snowplow based phoenix_events table.

sheyd:
#### What's this PR do?
* Added `phoenix_events` and `phoenix_sessions` tables to Quasar based on Snowplow.

#### How to test?
Ran this locally from my laptop against the Prod database with updated table names, both for create and refresh options, and everything looks good on the SQL side.

https://www.pivotaltracker.com/story/show/165548869


data/sql/derived-tables/snowplow_events.sql
```diff
@@ -72,3 +74,87 @@ CREATE TABLE public.snowplow_raw_events AS
...
+        e.modal_type,
+        e.session_id,
+        e.browser_size,
+        e.northstar_id,
+        e.device_id
+    FROM public.snowplow_raw_events e
+    LEFT JOIN public.campaign_info i ON i.campaign_id = e.campaign_id::bigint
+);
+CREATE UNIQUE INDEX spe_unique ON public.snowplow_phoenix_events (event_datetime, event_name, event_id);
+CREATE INDEX spe_session_id ON public.snowplow_phoenix_events (session_id);
```
sohaibhasan: Let's also add an index on timestamp

mjain-1: the unique index is on `event_datetime`, we deprecated `timestamp`...

mjain-1: wait sorry original comment is there's an index on `event_datetime`-- this should optimize all date-based queries. we no longer have a `timestamp` column!

PR #785, opened 2019-07-02
Adding phoenix_events and phoenix_sessions combined table sql.

sheyd:
#### What's this PR do?
* Adds SQL for combining Puck and Snowplow respective `phoenix_events` tables. 
https://github.com/DoSomething/quasar/compare/puck-sp-combined-events?expand=1#diff-7ddb641c696addd4d88b82589b3fbb65
Already manually run multiple times in Prod for Snowplow QA'ing. 
https://www.pivotaltracker.com/story/show/165548869


data/sql/derived-tables/snowplow_events.sql
```diff
@@ -158,3 +158,80 @@ CREATE TABLE public.snowplow_sessions AS (
...
 GRANT SELECT ON public.snowplow_sessions TO dsanalyst;
+
+
+DROP TABLE IF EXISTS public.phoenix_events_combined;
+CREATE TABLE public.phoenix_events_combined AS (
+    SELECT
+        p.event_id,
+        p.event_datetime,
+        p.event_name,
+        NULL as event_source,
```
mjain-1: Don't we have this field in the current `phoenix_events` table?

sheyd: Yeah, we do... not sure why I null'd it out. 🤔

sheyd: Should I update the `phoenix-next` events in the snapshot to be `phoenix` for consistency across the combined table, or do we want to keep source data as unchanged as possible?

mjain-1: hm, good question. is there a reason we made the jump from `phoenix` to `phoenix-next`? my initial thought is let's preserve source data, particularly because i don't think we're using this field anywhere in Looker. but let me think a little bit more if there are any other implications!

data/sql/derived-tables/snowplow_events.sql
```diff
@@ -25,6 +25,13 @@ CREATE INDEX base_event_id ON public.snowplow_base_event (event_id);
 GRANT SELECT ON public.snowplow_base_event TO dsanalyst;
 
 
+DELETE FROM public.snowplow_base_event
```
mjain-1: is there a benefit to doing it this way rather than during the above table creation?

data/sql/derived-tables/snowplow_events.sql
sheyd: Good news is, we can leave the naming as is, and the update to the source table is a simple SQL one, so we can always merge in and include the field (not as a `NULL` value), and update after the fact. I'll make the change to include it from the source table now, and we can decide on exact data naming later.

mjain-1: 👍 yeah i'm okay with leaving as is for now!

data/sql/derived-tables/snowplow_events.sql
sheyd: I think there's some ease of understanding and computationally efficiency to deleting a relatively small number of rows from the base table vs putting in an additional `where` clause. But I'm not tied to this, and the PostgreSQL query optimizer likely does a better optimization job than my manual separating out of steps. I can modify the table creation if you think that's a better choice!

data/sql/derived-tables/snowplow_events.sql
sheyd: Code updated here: https://github.com/DoSomething/quasar/pull/785/files#diff-7ddb641c696addd4d88b82589b3fbb65R176!

data/sql/derived-tables/snowplow_events.sql
mjain-1: got it! my initial thought was to use a `where not exists` but performance seems to be similar, so i'm fine with leaving as is!

mjain-1:
adding a reminder to rebase onto master to pull in adding event_id to sessions table!

data/sql/derived-tables/phoenix_events.sql
```diff
@@ -91,3 +91,50 @@ CREATE MATERIALIZED VIEW :phoenix_events AS (
 CREATE UNIQUE INDEX ON :phoenix_events (event_id, event_name, ts, event_datetime, northstar_id, session_id);
 GRANT SELECT ON :phoenix_events TO looker;
 GRANT SELECT ON :phoenix_events TO dsanalyst;
+
+DROP MATERIALIZED VIEW IF EXISTS :phoenix_sessions CASCADE;
+CREATE MATERIALIZED VIEW :phoenix_sessions AS (
+       SELECT
+		e.page #>> '{sessionId}' AS session_id,
+		min(e._id) as event_id,
```
mjain-1: is `event_id` sequential in phoenix?

sheyd: I don't think so. I thought you mentioned this morning this needing to have `min` value, but I could have misunderstood. Should it just be `event_id`? Or does it need the same logic [you used here](https://github.com/DoSomething/quasar/pull/785/files#diff-7ddb641c696addd4d88b82589b3fbb65R136)?

mjain-1: oh, I think I had mentioned using `first_value` similar to the landing page subquery

data/sql/derived-tables/phoenix_events.sql
```diff
@@ -91,3 +91,50 @@ CREATE MATERIALIZED VIEW :phoenix_events AS (
...
+		max(e.page #>> '{referrer, query, from_session}') AS from_session,
+		max(e.page #>> '{referrer, query, source}') AS referrer_source,
+		max(e.page #>> '{query, utm_source}') AS utm_source,
+		max(e.page #>> '{query, utm_medium}') AS utm_medium,
+		max(e.page #>> '{query, utm_campaign}') AS utm_campaign,
+		max(e1.landing_path) AS landing_page
+	FROM :events e
+	LEFT JOIN (
+		SELECT e.page #>> '{sessionId}' AS session_id,
+		FIRST_VALUE(e.page #>> '{path}') OVER (
```
mjain-1: I think this is where we'll want to grab the first `event_id` using similar logic

sheyd: Added full updated query [here](https://github.com/DoSomething/quasar/pull/785/files#diff-694653231036bb6587197aa9f561dd42R96). I'm running to see if it runs successfully. Small sample test seemed to work well!

data/sql/derived-tables/phoenix_events.sql
```diff
@@ -95,49 +95,51 @@ GRANT SELECT ON :phoenix_events TO dsanalyst;
...
-		SELECT e.page #>> '{sessionId}' AS session_id,
-		FIRST_VALUE(e.page #>> '{path}') OVER (
-		        PARTITION BY e.page #>> '{sessionId}'
-			ORDER BY (e.meta #>> '{timestamp}')::numeric)
-			AS landing_path
-		FROM :events e) e1
-	ON e1.session_id = e.page #>> '{sessionId}'
-	GROUP BY e.page #>> '{sessionId}'
+                e.page #>> '{sessionId}' AS session_id,
+                max(e1.event_id),
```
mjain-1: think we'll need to name this field with AS `event_id`, otherwise it'll be labeled `max`

sheyd: You're 💯 right, thanks for the catch!!!

data/sql/derived-tables/snowplow_events_refresh.sql
```diff
@@ -181,4 +193,92 @@ TRUNCATE public.snowplow_sessions;
...
+DROP TABLE IF EXISTS public.snowplow_sessions_stage;
+
+
+DROP TABLE IF EXISTS public.phoenix_events_combined_stage;
+CREATE TABLE public.phoenix_events_combined_stage AS (
+    SELECT
+        p.event_id,
+        p.event_datetime,
+        p.event_name,
+        NULL as event_source,
```
mjain-1: think this needs to be updated as well since we have this data in puck

sheyd: Updated!

PR #789, opened 2019-07-10
Syntax fix for duplicate utm* fields instead of referrer*

sheyd:
## What's this PR do?
* Fixes duplicate `utm_*` field references to be `referrer_*` fields in Snowplow `phoenix_events_stage` tables.
https://github.com/DoSomething/quasar/compare/referrer-fix?expand=1#diff-f24795805c95a44c7027d7660ac3eb7dR213


sheyd:
Also snuck in an `unlogged` tables statement for the staging tables, since found out from the an article and the PostgreSQL documentation:

>UNLOGGED
If specified, the table is created as an unlogged table. Data written to unlogged tables is not written to the write-ahead log (see Chapter 30), which makes them considerably faster than ordinary tables. However, they are not crash-safe: an unlogged table is automatically truncated after a crash or unclean shutdown. The contents of an unlogged table are also not replicated to standby servers. Any indexes created on an unlogged table are automatically unlogged as well.

PR #790, opened 2019-07-10
Adding utm params to users source details

blisteringherb:
#### What's this PR do?
This PR adds the `utm_medium`, `utm_source`, and `utm_campaign` fields to public.users source details. It's a slight modification to the code in the ticket.
You can test with something like this against the public.users table:
```
SELECT 
    source_detail,
    substring(source_detail from '(?<=utm_medium\:)(.*)(?=\,)') AS utm_medium,
    substring(source_detail from '(?<=utm_source\:)([^\,]*)') AS utm_source,
    substring(source_detail from '(?<=utm_campaign\:)(.*)(?=)') AS utm_campaign
FROM northstar.users
WHERE created_at>= now()-INTERVAL '1 day'
```
https://www.pivotaltracker.com/n/projects/2076969/stories/166626127


data/sql/derived-tables/users_table.sql
```diff
@@ -54,9 +54,9 @@ CREATE MATERIALIZED VIEW public.users AS
 		email_status."timestamp" AS cio_status_timestamp,
 		u.sms_status,
 		u.source_detail,
-		substring(u.source_detail from '(?<=utm_medium\:)(.*)(?=\,)') AS utm_medium,
-		substring(u.source_detail from '(?<=utm_source\:)([^\,]*)') AS utm_source,
-		substring(u.source_detail from '(?<=utm_campaign\:)(.*)(?=)') AS utm_campaign
+		substring(source_detail from '(?<=utm_medium\:)([\w*]+)') AS utm_medium,
```
mjain-1: I don't think we need `[]` in these cases. Also not sure if we need the `+` operator for `utm_medium`?

PR #792, opened 2019-07-15
Adding feature flags json to northstar.users and public.users

blisteringherb:
#### What's this PR do?
This PR pulls in the `feature_flags` object into a column on the northstar.users table. It's stored as a `varchar` and then put into a flat structure in the public.users mat view. 
I've added the necessary `feature_flags` column in QA so it can be tested there. 

This snippet can be used to test after running the `northstar_backfill` command to update the northstar.users table
```
CREATE MATERIALIZED VIEW public.users AS 
	SELECT 
		u.feature_flags::jsonb #>> '{badges}' AS badges
	FROM northstar.users u
	;
```
https://www.pivotaltracker.com/n/projects/2076969/stories/165832569


mjain-1:
@sheyd mind reviewing this? we'll want to have this ready for when badges launches in the next week or so!

data/sql/derived-tables/users_table.sql
```diff
@@ -54,6 +54,7 @@ CREATE MATERIALIZED VIEW public.users AS
 		email_status."timestamp" AS cio_status_timestamp,
 		u.sms_status,
 		u.source_detail,
+		u.feature_flags::jsonb #>> '{badges}' AS badges
```
sheyd: @blisteringherb I don't want to add work to what you've already setup, but if `feature_flags` is coming in as a JSON blob, would it make more sense to set the value of the column to `jsonb` instead of `varchar`?

sheyd:
@blisteringherb Looks like you might also have to rebase off of master for the merge conflicts for the `utm_*` field updates to the users table.

mjain-1:
@blisteringherb let me know what the LOE is for this / want to make sure Jen and/or Katie are aware of when we'll be ready to capture data!

mjain-1:
@blisteringherb looks like this needs to rebased? and then is this ready to be merged (cc @sheyd)?

PR #797, opened 2019-08-01
Checking for curly brace at EOS to avoid truncated values

blisteringherb:
#### What's this PR do?
This modifies the json string test so that its searching at the end of the string. This is somewhat of a hack, and we could also add logic to parse the broken strings with regex, but I believe the issue affects only a handful of signups and it source issue in Rogue should be rectified quickly so we shouldn't have this issue in the future.

https://dosomething.slack.com/archives/C9P152E5R/p1564616762000100
https://www.pivotaltracker.com/n/projects/2076969/stories/167649973

### DIFF
--- a/data/sql/derived-tables/campaign_activity.sql
+++ b/data/sql/derived-tables/campaign_activity.sql
@@ -15,15 +15,15 @@ CREATE MATERIALIZED VIEW public.signups AS
         sd.created_at AS created_at,
         sd.source_details,
         CASE 
-			WHEN source_details ILIKE '\{%%' 
+			WHEN source_details ILIKE '%%\}'
 			THEN (CAST(source_details as json) ->> 'utm_medium') 
 			ELSE NULL END AS utm_medium,
 		CASE 
-			WHEN source_details ILIKE '\{%%' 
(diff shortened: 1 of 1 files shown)


sheyd:
Pending @mjain-1 approving this will lead to some unaccounted for lost values, approved!

data/sql/derived-tables/campaign_activity.sql
```diff
@@ -15,15 +15,15 @@ CREATE MATERIALIZED VIEW public.signups AS
         sd.created_at AS created_at,
         sd.source_details,
         CASE 
-			WHEN source_details ILIKE '\{%%'
```
mjain-1: does changing this not affect previous values??

mjain-1:
thanks!! do we know how many rows this affects?

sheyd:
@mjain-1 As of last night, it was only 1 row I found with the truncated value, but when we run this, there might be more.

sheyd:
We can do a simple query looking for `[` in the `source_details` field.

mjain-1:
cool, yeah might be helpful to have and document!

PR #798, opened 2019-08-02
Adding dbt to repo

blisteringherb:
#### What's this PR do?
This PR adds the DBT code to the Quasar repo so that everyone can start working on their respective parts with the framework in place.

https://www.pivotaltracker.com/n/projects/2076969/stories/167621981


sheyd:
@blisteringherb If we want to have the folder that contains `dbt` files to be names `dbt`, we have to [change the name](https://github.com/DoSomething/quasar/pull/798/files#diff-9464235ad31ae62317534b9ac5104902R5) to `dbt`. 

I have no particular preference, but I don't think the dbt commands will work if the naming is off.

PR #799, opened 2019-08-05
feature: Expose carrier_delivered_at and carrier_failure_code in gambit_messages_outbound

rapala61:
#### What's this PR do?
- It adds the `carrier_delivered_at` and `carrier_failure_code` columns to the `gambit_messages_outbound` derived table.

### DIFF
--- a/data/sql/derived-tables/gambit_messages.sql
+++ b/data/sql/derived-tables/gambit_messages.sql
@@ -1,6 +1,6 @@
 DROP MATERIALIZED VIEW IF EXISTS :ft_gambit_messages_flattened CASCADE;
 CREATE MATERIALIZED VIEW :ft_gambit_messages_flattened AS
- (SELECT 
+ (SELECT
     agent_id AS agent_id,
     attachments->0->>'contentType' AS attachment_content_type,
     attachments->0->>'url' AS attachment_url,
@@ -13,6 +13,8 @@ CREATE MATERIALIZED VIEW :ft_gambit_messages_flattened AS
     macro AS macro,
     match AS match,
     metadata #>> '{delivery,queuedAt}' AS delivered_at,
+		metadata #>> '{delivery,failedAt}' as failed_at,
+    metadata #>> '{delivery,failureData,code}' as failure_code,
     (metadata #> '{delivery}' ->> 'totalSegments')::INT AS total_segments,
     platform_message_id as platform_message_id,
     template AS template,
@@ -86,6 +88,9 @@ CREATE MATERIALIZED VIEW public.gambit_messages_outbound AS
 	f.message_id,
 	f.macro,
 	f."match",
+	f.delivered_at,
+	f.failed_at,
+	f.failure_code,
 	f.platform_message_id,
 	f.template,
 	f.text,
@@ -106,6 +111,9 @@ CREATE MATERIALIZED VIEW public.gambit_messages_outbound AS
 	g.message_id,
 	g.macro,
 	g."match",
+	g.delivered_at,
+	g.failed_at,
+	g.failure_code,
 	g.platform_message_id,
 	g.template,
 	g.text,
@@ -125,6 +133,6 @@ CREATE MATERIALIZED VIEW public.gambit_messages_outbound AS
 	g.direction <> 'inbound'
 	AND g.user_id IS NULL));
 
-CREATE INDEX ON public.gambit_messages_outbound (message_id, created_at, user_id, conversation_id);
+CREATE INDEX ON public.gambit_messages_outbound (message_id, created_at, delivered_at, user_id, conversation_id);
 GRANT SELECT ON public.gambit_messages_outbound to looker;
 GRANT SELECT ON public.gambit_messages_outbound to dsanalyst;
--- a/data/sql/derived-tables/gambit_messages.sql
+++ b/data/sql/derived-tables/gambit_messages.sql
@@ -12,11 +12,8 @@ CREATE MATERIALIZED VIEW :ft_gambit_messages_flattened AS
     _id AS message_id,
     macro AS macro,
     match AS match,
-    metadata #>> '{delivery,queuedAt}' AS sent_at,
-		metadata #>> '{delivery,deliveredAt}' AS delivered_at,
-	  metadata #>> '{delivery,failedAt}' as failed_at,
-    metadata #>> '{delivery,failureData,code}' as failure_code,
-    (metadata #>> '{retryCount}')::INT AS retry_count,
+		metadata #>> '{delivery,deliveredAt}' AS carrier_delivered_at,
+    metadata #>> '{delivery,failureData,code}' as carrier_failure_code,
     (metadata #> '{delivery}' ->> 'totalSegments')::INT AS total_segments,
     platform_message_id as platform_message_id,
     template AS template,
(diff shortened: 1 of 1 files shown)


data/sql/derived-tables/gambit_messages.sql
```diff
@@ -125,6 +133,6 @@ CREATE MATERIALIZED VIEW public.gambit_messages_outbound AS
 	g.direction <> 'inbound'
 	AND g.user_id IS NULL));
 
-CREATE INDEX ON public.gambit_messages_outbound (message_id, created_at, user_id, conversation_id);
+CREATE INDEX ON public.gambit_messages_outbound (message_id, created_at, delivered_at, user_id, conversation_id);
```
mjain-1: what information does `delivered_at` provide, and does it need to be included in this index?

rapala61: The `delivered_at` property holds the time when we receive a delivery confirmation receipt from Twilio to Blink. A few months ago, I identified cases when carriers take hours to actually deliver (technically this could also mean some carriers take long to confirm delivery) messages to members. If we ever need to run any analysis about that, we would need this property. I don't think is imperative to index it now. I just did because I exposed the property and if we needed to use it, we could efficiently query against it.

mjain-1: ah okay, maybe we can update the field name to be `carrier_delivered_at` for clarity. 

this index needs to be updated generally, but maybe instead of adding this field here we could create a 'error deliverability' index that includes (not sure what order we'd want) `failure_code`, `created_at`, `delivered_at` or something

rapala61: @mjain-1 I just realized now that we are calling `deliverd_at` to what really should be `sent_at`. I'm renaming it to `sent_at`. Do you think that with this change `delivered_at` should still be changed to `carrier_delivered_at`?

Could you point me to an example of a named index like the one you suggested? I though I just had to add the column to the CREATE INDEX line.

sheyd: I'm very into @mjain-1 idea of having a separate index for tracking deliverability. This table's indexes are ones I'll be taking on as part of performance improvement work next sprint. As a starting point, let's do `carrier_delivered_at` which includes `failure_code`, `created_at`, and `delivered_at`. I'm tempted to do `failed_at` as well to track on-going failures, but I think that's less important than seeing if a message got delivered at all, what failure codes it had, and the time-delta on failures from creation to delivery.

sheyd: @rapala61 regarding
>Could you point me to an example of a named index like the one you suggested? I though I just had to add the column to the CREATE INDEX line.

You can [name an index](https://www.postgresql.org/docs/11/sql-createindex.html) as an option but it's not a requirement in PostgreSQL. If you omit the index name, one will be created automatically.

rapala61: Ah got it, thanks @sheyd . So we want to crate a separate index like:

```
CREATE INDEX deliverability ON public.gambit_messages_outbound (created_at, sent_at, delivered_at, failed_at, failure_code);
```

mjain-1: @rapala61 
> Do you think that with this change delivered_at should still be changed to carrier_delivered_at? 

i think `sent_at` might still be confusing given there is `created_at` (which i think is when we sent it, right?), so maybe we can do `carrier_sent_at`?

> CREATE INDEX deliverability ON public.gambit_messages_outbound (created_at, sent_at, delivered_at, failed_at, failure_code); 

agreed with @sheyd that we probably don't need all the time columns in the index to begin with. i think the most common query will be knowing how many error codes /  month so maybe something along the lines of `ON created_at, failure_code` to start. then, if we find we are doing a lot of analyses based on `failed_at` or `sent_at` we can add those...

rapala61: > i think sent_at might still be confusing given there is created_at (which i think is when we sent it, right?), so maybe we can do carrier_sent_at?

`created_at` will always be set for all messages (inbound or outbound). `sent_at` (`queued_at` in the raw Gambit table) would only be set if/when the message was actually sent to Twilio for delivery (not the carrier). This is why is also a little confusing to me to call it `carrier_sent_at`. Unlike, `carrier_delivered_at` which is indeed a timestamp reported by Twilio based on the carrier.

I will change `delivered_at` to `carrier_delivered_at` which definitely conveys the right meaning. Thanks for that suggestion! @mjain-1 

What about this INDEX? @sheyd @mjain-1 

```
CREATE INDEX deliverability ON public.gambit_messages_outbound (created_at, failure_code);
```

sheyd: @rapala61 
> What about this INDEX? @sheyd @mjain-1

Sounds good to me!

rapala61:
@sheyd @mjain-1 The PR is ready for one final 👀. Thanks for reviewing before!

data/sql/derived-tables/gambit_messages.sql
```diff
@@ -106,6 +119,11 @@ CREATE MATERIALIZED VIEW public.gambit_messages_outbound AS
 	g.message_id,
 	g.macro,
 	g."match",
+  g.sent_at,
```
mjain-1: i'm still finding this to be kind of confusing terminology that doesn't make it clear when to use `created_at` vs. `sent_at` especially given that we've typically relied on `created_at` to more or less be defined as `sent_at`. 

to make sure I understand, the benefit of having both `sent_at` and `delivered_at` is we can know where the point of failure is, twilio or the carrier?

rapala61: @mjain-1 I'll defer to you as far as what to do with the `sent_at`. It's not important in order to get the error codes to looker which is what this PR really is all about!.

The importance of `carrier_delivered_at` is that we can detect messages that carriers are not reporting as delivered or not. We can also detect messages that carriers report to have delivered hours or days later that the date we sent them.

Let me know what the final decision is on `created_at`, I will push the change and hopefully merge this PR. :)

mjain-1: totally; I guess I'm wondering if there's overlap between the info that those additional columns provide and the error code? and then if we do think these are things we'll be regularly querying, would it make more sense to think about a separate `gambit_error_log` table rather than adding to this one which we generally use to under broadcast response rates etc.? cc @sheyd curious what your thoughts are

sheyd: I think I'd prefer to keep things as clear as possible, but include more information than less for this iteration (like having the extra column, but not necessarily indexing on it if we don't need to query it for now).

My thoughts are, let's prepend anything that is not internal to our systems with `carrier_`, and use names we use in other DS systems for consistency. So something like:
- `created_at` - When Gambit actually created the message
- `sent_at` - When Gambit submitted to Twilio
- `carrier_sent_at` - When Twilio sent the message to a number
- `carrier_delivered_at` - When Twilio got delivery confirmation of the message being sent to the number.

I think I'm over-counting the number of fields we're talking about here, but does this approach make sense?

I think it's useful to talk about the merits of a `gambit_error_log` table, since us having centralized analytics I think is important for business reasons. That being said, I'd like to make sure we don't use Quasar as a de-bugging service for other systems, but as an indicator of when things are impacting our critical KPI's (including revenue). In this case, since we've seen over the past year that message deliverability has been a bit of a black box, I think it would be great to have a top-level deliverability report on SMS, but mostly as a roll-up table. De-bug handling should ideally take place in the source system it originates in via good logging, which I'm pretty sure Gambit has already.

We can discuss more in Sprint planning.

PR #800, opened 2019-08-07
Gdpr removal

sheyd:
#### What's this PR do?
* Anonymizes Northstar and CIO GDPR user data provided via CSV file.
https://github.com/DoSomething/quasar/pull/800/files#diff-a0269408c633c68be7b513ba97ebc07f

Tested with test tables in Quasar Prod, and confirmed removes records appropriately.

https://www.pivotaltracker.com/story/show/167624607


sheyd:
@mjain-1 This is for the GDPR compliance removal of user data from Northstar and CIO tables. I've used the [Northstar](https://github.com/DoSomething/northstar/blob/master/app/Console/Commands/DeleteUsersCommand.php#L55-L59) code as a reference for what to preserve. I've also used it to inform CIO fields to remove/retain. If you could take a look to make sure the fields are in-line with what we want to remove/retain, I'd appreciate it!

data/sql/misc/gdpr_northstar_removal.sql
```diff
@@ -0,0 +1,27 @@
+UPDATE :users
+	SET birthdate = date_trunc('year', birthdate)
+	WHERE id = ':nsid';
+
+UPDATE :users
```
mjain-1: looking at the previously GDPR'ed accounts + this list, can we keep these values:
`sms_status`, `last_accessed_at`, `last_authenticated_at`, `last_messaged_at`, `voter_registration_status`, `email_subscription_status`. doesn't look like they'd be PII based on this information!

data/sql/misc/gdpr_cio_removal.sql
```diff
@@ -0,0 +1,35 @@
+UPDATE :customer_event
```
mjain-1: do these fields `template_id`, `email_id`, `event_id`, `subject`, `href`, `link_id` contain PII?

data/sql/misc/gdpr_northstar_removal.sql
sheyd: For this I want to be consistent with how Northstar handles field removal, and basically it deletes/nulls out everything but [these fields](https://github.com/DoSomething/northstar/blob/master/app/Console/Commands/DeleteUsersCommand.php#L55-L59). If you think there's historical analytics value to the data, definitely happy to re-consider, since we might get value out of them, but Northstar might not. 

Let's review the GDPR compliance on those fields with @mshmsh5000, since he's our resident expert.

data/sql/misc/gdpr_cio_removal.sql
sheyd: Probably not, but I'm opting for deleting most things that aren't Northstar ID and some metadata about the event, and timestamp. Happy to talk thru these individually as well.

mjain-1: I can see us wanting `template_id` as that relates to specific newsletters.

Between email_id/event_id, is one a primary key at the moment? 

Subject, href, link_id: sure, we can get rid of these as I guess we'll be able to get that info from `template_id`.

data/sql/misc/gdpr_northstar_removal.sql
mjain-1: Hm, yeah those fields don't seem to be included in what's being dropped in Northstar?

mjain-1: Or are those the only fields that we're keeping? hah sorry!

sheyd: Yep! I was confused when I first read the code too, those are the fields that are kept!

data/sql/misc/gdpr_cio_removal.sql
sheyd: Got it, I'll update to keep `template_id`. `email_id` is part of the current primary key.

data/sql/misc/gdpr_northstar_removal.sql
mjain-1: got it got it, I think the only one that I would push to maintain is the `voter_registration_status` to understand historically how many we've registered, etc.

sheyd: @mshmsh5000 Thoughts on `voter_registration_status` field falling under GDPR or not?

data/sql/misc/gdpr_cio_removal.sql
mjain-1: ah okay, we probably don't want the primary key to contain null values, and I think we're not getting any PII from that anyway so should be fine to not null out.

data/sql/misc/gdpr_northstar_removal.sql
sheyd: Check IRL with Matt, and we can keep `voter_registration_status`.

data/sql/misc/gdpr_cio_removal.sql
sheyd: @mjain-1 I think with keeping `email_id` this is ready to merge in?

mjain-1: think so!

mjain-1: not sure if you required Dave's review but will approve

PR #817, opened 2019-09-13
Pipenv

sheyd:
#### What's this PR do?
* Updates code to use [Pipenv[(https://docs.pipenv.org/en/latest/) instead of `requirements.txt` file for dependency tracking. I made sure all dependencies from `requirements.txt` were imported to Pipenv.
* Removes `requirements.txt`.
* Updates documentation with `pipenv` commands, removed Vagrant section for local database dev, and expands on linting rules.
https://github.com/DoSomething/quasar/compare/pipenv?expand=1#diff-1e61a31bf9b94805f869dc4137ec1885
I manually ran a local install and build with `pipenv` with no issues. Before merge to master, I'll update the Jenkins QA `Deploy Branch` job to use `pipenv` commands and see if it works well.


blisteringherb:
Does this need changes to `setup.py` since that references `requirements.txt?

sheyd:
Hmm, yeah it does, and there's some [guidance for it](https://docs.pipenv.org/en/latest/advanced/#pipfile-vs-setup-py). Reading up and seeing how to best approach this.

sheyd:
So, there's a lot of back and forth about this... I'm wondering, based on the Pipenv guidance above, if we should remove the `install_requires` directive from the `setup.py` since we're not setting up a library project, but an application, and as long as we have the `pipenv install` step run, it should build with no issues, since the virtualenv will have all the necessary dependencies installed. Thoughts?

rapala61:
I setup everything using the old instructions. Any cleanup steps I should take before re-running the setup with pipenv?

sheyd:
@rapala61 No, although pause on testing, since for some reason my local build went okay without `requirements.txt` in place. Re-testing.

sheyd:
@blisteringherb [Based on this issue] I borrowed to code to generate a new `requirements` list from Pipenv, which seems to work. We can use this to keep parity with how we've been doing things for now and debate keeping the `install_requires` in a separate issue.

sheyd:
@rapala61 If you pull down the latest changes, it should hopefully work. I would clone to a new folder and see if the `Pipenv` install instructions work for you in a clean slate (and I appreciate the thorough testing a lot! 🏆), and then move your dev over there instead of the legacy folder.

README.md
```diff
@@ -135,13 +105,20 @@ https://docs.python.org/3/distutils/setupscript.html
...
-make test
-```
+We currently don't have any tests setup. :frowning:
 
 ### End to end tests
 
 ### Coding style tests
+Multiple options are available here, but usually we stick to PEP8 syntax checking. 
+You can set one up in your editor/IDE of choice.
+If you like the stick to the CLI or run a manual check,
```
rapala61: Seems like there's a small typo: "If you like `to` stick to the CLI"

sheyd: Updated!

sheyd:
> Also, should we add some steps to setup and test linting in at least 1 or two most popular IDEs like VSCode?

I think this is a great idea, and something to put in a subsequent PR. I'd recommend Sublime and VS Code, since I think that covers most of our use-cases, but we can file a ticket to update it!

PR #821, opened 2019-09-19
Add query params to Snowplow derived tables in custom SQL and DBT.

sheyd:
#### What's this PR do?
* Adds URL query parameters from raw Snowplow events to web analytics tables to both current SQL and DBT code.
https://github.com/DoSomething/quasar/compare/snowplow-query-params?expand=1#diff-7ddb641c696addd4d88b82589b3fbb65R11
Will push to QA to test.
https://www.pivotaltracker.com/story/show/168652309


data/sql/derived-tables/snowplow_events.sql
```diff
@@ -55,6 +56,7 @@ CREATE TABLE public.snowplow_raw_events AS
 	b.event_type,
 	b."host",
 	b."path",
+    b.query_parameters,
```
blisteringherb: Perhaps a silly question, should this be aligned?

sheyd: Not a silly question at all. Linting 👨‍⚖ definitely welcome.

data/sql/derived-tables/snowplow_events.sql
```diff
@@ -94,6 +94,7 @@ CREATE TABLE public.snowplow_phoenix_events AS (
         e."path",
         e."host",
         e.query_parameters,
+        e.url as event_url,
```
mjain-1: not sure this is the right field... is this from the payload for `phoenix_clicked_link_action` events?

mjain-1:
actually sorry let's rename the field from `url` to `link_event_url` or something a little more clearer to distinguish from the page url!

data/sql/derived-tables/snowplow_events.sql
sheyd: The `url` field gets populated [from here](https://github.com/DoSomething/quasar/pull/821/files#diff-7ddb641c696addd4d88b82589b3fbb65R40). If the `url` field exists in the `payload` column in the `ft_snowplow.snowplow_event` table, then it's propagated thru, otherwise it's NULL. I might be misunderstanding where this `url` field should be coming from?

mjain-1: yeah sorry i think i was confused by the naming because there is page_url and then there's what you've described... think updating  the name to be clearer about the fact that this is coming from an event and not the page the user is on will help

sheyd:
@mjain-1 Superb idea, since `url` was confusing when I went back thru the code and pushed the PR yesterday 😂. Changing in a minute!

PR #846, opened 2019-10-25
Newsletter table fixes

mjain-1:
#### What's this PR do?
Collecting updates to the table to confirm it's accurate:

- A user unsubscribing from a topic should trump global unsubscribe (e.g. a check for if a user unsubs from a topic before unsubscribing completely) 

user_newsletter_subscriptions.sql

- Confirm that topic_unsubscribed_at reflects the topic unsubscribed at date if it's earlier than the newsletter_unsubscribed_at date


mjain-1:
@blisteringherb do we need to rebuild for this to take effect?

blisteringherb:
@mjain-1 Because we're using the "table" strategy for DBT, this should just rebuild during the next run and will incorporate these changes.

PR #865, opened 2019-11-05
Updating campaign info international reference

blisteringherb:
#### What's this PR do?
This updates the incorrect reference in the campaign_info_international build.


sheyd:
@blisteringherb UGH, the doc diffs have hit us hard here. 😒 

Might be worth re-doing another PR off of current `master` branch so doc diffs don't occur.

PR #875, opened 2019-11-11
Updates DBT docs to use Prod schema but with schema only access user.

sheyd:
#### What's this PR do?
Updates DBT docs to to use Prod schema based on a read-only schema access user so we don't have odd schema's like `dbt_schema` and our doc diffs aren't so crazy between different devs updating them.

Setup the DBT Docs user profile in your `profiles.yml` file.

Checkout branch and run `dbt docs generate` using the `docs` profile and see if you get any git diffs!

If no diffs, things are working, but make a final inspection via `dbt docs serve` to make sure everything looks good.

https://www.pivotaltracker.com/story/show/169357116


rapala61:
@sheyd what do you think if we add a sample `profiles.yml`? We do this in [Gambit](https://github.com/DoSomething/gambit/blob/master/.env.example). I think it would be helpful here. We can use dummy credentials and replace locally with what's in LastPass.

sheyd:
@rapala61 Great idea! [Added](https://github.com/DoSomething/quasar/pull/875/files#diff-6a560fee8346f1f1c523d28c18280f4a).

PR #879, opened 2019-11-13
Recreating Phoenix events docs with doc blocks

blisteringherb:
#### What's this PR do?
This PR adds the DBT documentation for web events (Phoenix and Snowplow). The docs have been generated with the documentation generation profile.

Note:
`campaign_id` is already defined in the Gambit Messages doc block, and its definition is fairly specific to Gambit. Since we can only have one field definition at a time, we'll either need to make any definitions more general, or we can just use description strings as we are in this PR.


rapala61:
@blisteringherb What do you think of recompiling the docs so that they are built against the `public` schema instead?

sheyd:
I pushed up the commit to use the `public` schema for less diffs.

PR #881, opened 2019-11-13
Couple of small syntax & linting updates to MEL code before rolling out.

sheyd:
#### What's this PR do?
Provides some small syntax and naming updates to MEL:
* I kept the alias `mel` for internal dependency references, but the model being named `mel` was not very descriptive and very institutional knowledge biased.
* Fixed two instances of lower-cased SQL syntax operators that were bugging me.
https://github.com/DoSomething/quasar/compare/mel-lint?expand=1#diff-9464235ad31ae62317534b9ac5104902R203
Will test in QA when running MEL migration in a few days.
Was reviewing current MEL SQL with DBT model SQL and tracked down the syntax issues.
https://www.pivotaltracker.com/story/show/169683230


blisteringherb:
Do you want to regenerate the docs so that the name of the model gets updated?

sheyd:
Ohhh, good call!

PR #882, opened 2019-11-13
Fixes reference in user_activity_dbt_validation.sql

rapala61:
#### What's this PR do?
Fixes reference to `public.users`. It should reference `ds_dbt.users` to test the users DBT pipeline dependency of the user_activity view.


sheyd:
@rapala61 We need to update references in dbt-validation MEL as well, right?

sheyd:
[Here](https://github.com/DoSomething/quasar/blob/master/data/sql/derived-tables/mel_dbt_validation.sql#L147) and [here](https://github.com/DoSomething/quasar/blob/master/data/sql/derived-tables/mel_dbt_validation.sql#L152) if I'm not mistaken.

rapala61:
@sheyd you're right! Thanks for pointing that out!

PR #890, opened 2019-11-14
Fix index syntax for MEL model.

sheyd:
#### What's this PR do?
Another index syntax issue I forgot to catch! 
https://github.com/DoSomething/quasar/compare/mel-index-fix?expand=1#diff-9464235ad31ae62317534b9ac5104902R208


setup.py
```diff
@@ -7,7 +7,7 @@
 
 setup(
     name="quasar",
-    version="2019.11.14.1",
+    version="2019.11.14.2",
```
rapala61: Is this the only change in this PR?

sheyd:
This is weird... trashing the PR.

PR #903, opened 2019-11-21
Adding school_id from northstar.users

blisteringherb:
#### What's this PR do?
This PR adds the `school_id` to the public.users table.
https://www.pivotaltracker.com/n/projects/2076969/stories/169718533


blisteringherb:
@sheyd Good call, I'll update that definition in the DBT users docs PR.

PR #919, opened 2019-12-16
Removing Rogue ack code since Rogue no longer sends data to Blink.

sheyd:
#### What's this PR do?
Since Rogue no longer [sends data to Blink](https://github.com/DoSomething/rogue/compare/7a062e0c...3aac536d), we can remove the simple Rogue event `ack` code.
[This code removed](https://github.com/DoSomething/rogue/compare/7a062e0c...3aac536d), whereas [this code](https://github.com/DoSomething/quasar/compare/rogue-queue-code-remove?expand=1#diff-9cfe7b6ae8cec24aa71005efd0fb7d6fR174-R202) remains, but looks like re-add due to moving indentation level.
Will test out in QA to ensure nothing fundamentally breaks.

https://www.pivotaltracker.com/story/show/170243897


sheyd:
This doesn't need to be merged/reviewed till 2020, staging the PR here so we can hit the ground running.

PR #938, opened 2020-01-31
Update broken field references in Northstar dependent tables

blisteringherb:
#### What's this PR do?
This PR fixes references to the field previously named `id` and points those to `northstar_id` since that's the new name of the field in the deduped layer. It also updates the address street field names.

https://www.pivotaltracker.com/n/projects/2076969/stories/171028589


quasar/dbt/models/member_event_log/member_event_log.sql
```diff
@@ -49,7 +49,7 @@ FROM {{ ref('posts') }} p
 WHERE p.status IN ('accepted', 'confirmed', 'register-OVR', 'register-form', 'pending')
 UNION ALL
 SELECT DISTINCT 
-    u_access.id AS northstar_id,
+    u_access.northstar_id AS northstar_id,
```
sheyd: Syntactically, you don't need `AS northstar_id` since the column is already named `northstar_id`.

quasar/dbt/models/member_event_log/member_event_log.sql
```diff
@@ -65,7 +65,7 @@ AND u_access.email IS DISTINCT FROM 'juy+runscopescheduledtests@dosomething.org'
 AND (u_access.email NOT ILIKE '%@example.org%' OR u_access.email IS NULL) 
 UNION ALL
 SELECT DISTINCT 
-    u_login.id AS northstar_id,
+    u_login.northstar_id AS northstar_id,
```
sheyd: Same as above:
>Syntactically, you don't need AS northstar_id since the column is already named northstar_id.

quasar/dbt/models/users_table/users.sql
```diff
@@ -1,5 +1,5 @@
 SELECT
-	u.id AS northstar_id,
+	u.northstar_id AS northstar_id,
```
sheyd: Another instance of
>Syntactically, you don't need AS northstar_id since the column is already named northstar_id.

PR #942, opened 2020-02-05
Unblock prod deploy

blisteringherb:
#### What's this PR do?
The Northstar Fivetran work is blocking production deployments. This PR removes the references to the new Fivetran tables and reverts field references to `id` when pulling data from the `northstar.users` table. Because the `northstar.users` source works fine, I'm using the source reference rather than the hardcoded value.

https://www.pivotaltracker.com/n/projects/2076969/stories/171105483


blisteringherb:
@rapala61 Thank you for the reminder, I just updated those two fields as well.

PR #944, opened 2020-02-05
Fixing reference to wrong field name

blisteringherb:
#### What's this PR do?
This fixes the one broken reference. For some reason locally the `users` model still isn't building and it mentions that `northstar_id` isn't found, but that field isn't referenced in that file. Trying to build it again now on Jenkins QA with the current build, but it takes quite a while.


docs/compiled/ds_dbt/member_event_log/member_event_log.sql
```diff
@@ -29,7 +29,7 @@ WHEN s."source" NOT LIKE '%sms%'AND s."source" NOT LIKE '%email%' AND s."source"
 WHEN s."source" ILIKE '%email%' THEN 'email'
 WHEN s."source" ILIKE '%niche%' THEN 'niche_coregistration'
 WHEN s."source" NOT LIKE '%sms%'AND s."source" NOT LIKE '%email%' AND s."source" NOT LIKE '%niche%' AND s."source" NOT IN ('rock-the-vote', 'turbovote') AND s."source" IS NOT NULL THEN 'other' END) AS "channel"
-FROM "quasar_prod_warehouse"."public"."signups" s
```
rapala61: Are the docs being generated with a right profile?

PR #954, opened 2020-02-18
User engagement newsletter

sheyd:
#### What's this PR do?
* Adds User Engagement E-mail SQL to data model.
https://github.com/DoSomething/quasar/compare/user-engagement-newsletter?expand=1#diff-9464235ad31ae62317534b9ac5104902R282
Will test in QA.
https://www.pivotaltracker.com/story/show/170958230
#### Questions:
@eriqa, can you validate that the `user_newsletter_cal_status` SQL you generated [here](https://drive.google.com/drive/u/0/folders/1BrvR0bpWRWvAm9SyLFls46D3pl7F-0yY?ths=true) is the equivalent of the SQL I wrote [here](https://github.com/DoSomething/quasar/compare/user-engagement-newsletter?expand=1#diff-1e423506c244d07f570abdcad36ccdce)?


blisteringherb:
Will we need indexes on any of these tables? It also looks like we might want to add some newlines to the ends of files.

sheyd:
I haven't tested out, but yeah... looking at the queries probably want to add some indexes at least on the `northstar_id` fields.

PR #977, opened 2020-03-18
Adds p.id in the grouping clause of the accepted reportbacks subquery

rapala61:
#### What's this PR do?
- It fixes the subquery to check if the reportback is approved by including the `p.id` column in the grouping by clause. Without it, multiple submissions by the ssame user in the same campaign id would not be included in the subquery and the reportback would not make it into the `reportbacks` table.
- https://www.pivotaltracker.com/story/show/171677933


rapala61:
The query was found to be working as expected after all. Resolution is included in the Pivotal ticket linked in the PR description.

PR #980, opened 2020-03-23
Updating TMC workflow to use more standard model.

sheyd:
#### What's this PR do?
* Reworks TMC data model to make `tmc_users_matched` file available as a copy in `public` schema instead of altering table ownership/permissions of a raw data table. This is more inline with how we source data from our raw layer in our DBT data model. It also cuts down on some TMC data pipeline errors altering table permissions causes. 
* Adds some more comments to TMC data model for clarity.
https://github.com/DoSomething/quasar/compare/tmc-model-rework?expand=1#diff-9464235ad31ae62317534b9ac5104902R253
QA!

https://www.pivotaltracker.com/story/show/171774534


quasar/dbt/dbt_project.yml
```diff
@@ -250,17 +250,22 @@ models:
...
           alias: tmc_users_out
           materialized: table
+          # This hook needed, since schema is moved in post-hook.
           pre-hook:
             - "DROP TABLE IF EXISTS tmc_out.tmc_users_out CASCADE"
           post-hook:
             - "ALTER TABLE {{ this }} SET SCHEMA tmc_out"
             - "GRANT SELECT ON ALL TABLES IN SCHEMA tmc_out TO tmc"
-            # Adds permissions for Looker and analyst access for processed TMC voter reg table.
+        tmc_users_matched:
```
rapala61: Do you think it's helpful to add some comment here to explicitly document this model is just a copy/paste of the `tmc_in.tmc_users_matched` table?

sheyd: @rapala61 Do you think a multi-line comment that briefly explains it is best here, or documenting the generation logic/lineage in the Pivotal ticket and then link the ticket here?

quasar/dbt/models/tmc_users/tmc_users_matched.sql
```diff
@@ -0,0 +1,2 @@
+SELECT *
+FROM {{ env_var('TMC_USERS_MATCHED') }}
```
rapala61: It was a bit confusing at first that the env var name is the same as the model. Do you think it would be clearer to name it something **like** `TMC_USERS_MATCHED_SOURCE`?

sheyd: Yep, let's definitely make that change!

quasar/dbt/dbt_project.yml
sheyd: Added some inline commentary to explain why we're doing this.

PR #988, opened 2020-04-06
Updates to user_newsletter_signups model to incorporate signup source and simplify periods to 1 month

eriqa:

### DIFF
--- a/quasar/dbt/models/user_newsletter_signups/user_newsletter_signups.sql
+++ b/quasar/dbt/models/user_newsletter_signups/user_newsletter_signups.sql
@@ -1,6 +1,7 @@
 WITH 
 recent_signups AS (
-	SELECT s.northstar_id, s.created_at AS signup_date, s.id
+	SELECT s.northstar_id, s.created_at AS signup_date, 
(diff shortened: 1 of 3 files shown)


sheyd:
@eriqa First off, _BRAVO_ on your first DBT PR! It's a joy that you've included docs and tests. 😺 

I'm going to checkout the branch since I think there's a small indentation issue, but can't tell from PR. Aside from that looks good, and I'll approve it shortly!

quasar/dbt/models/user_newsletter_signups/user_newsletter_signups.sql
```diff
@@ -1,6 +1,7 @@
 WITH 
 recent_signups AS (
-	SELECT s.northstar_id, s.created_at AS signup_date, s.id
```
sheyd, line 3: Double checking we don't need the `s.id` in the CTE anymore?

eriqa: I didn't think we needed it but now I'm second-guessing whether it maybe useful to keep it in case we event want to analyze the report-back associated with that signup. I can modify the branch and commit again. Let's just have a quick chat to understand the indentation issue to I don't bring it back in.

PR #990, opened 2020-04-06
voter-reg-quiz-funnel

sheyd:
#### What's this PR do?
* Adds the Voter Reg Quiz funnel SQL to DBT.
* Adds voter reg DBT model for future Voter Reg SQL.
https://github.com/DoSomething/quasar/compare/voter-reg-quiz-funnel?expand=1#diff-16ccfa5288341641cfee1c09e8b9fa6f
Will run in QA.

https://www.pivotaltracker.com/story/show/172157510


sheyd:
@rapala61 Tagging in @sohaibhasan, since I didn't do any of the SQL, so as long as it makes sense to him/anyone looking at voter reg, I'm okay with the names as they are. Although some `---` SQL comments would be nice so we have some guidelines. :)

rapala61:
Got it! Forgot that you were mostly integrating his SQL work. However, I do believe it's necessary to add some comments or update naming to make the SQL a bit more friendlier for anyone that might work on it in the future (I have no idea what those names mean :) ).

quasar/dbt/dbt_project.yml
```diff
@@ -318,6 +318,7 @@ models:
           alias: voter_reg_quiz_funnel
           materialized: table
           post-hook:
+            - "CREATE INDEX ON {{ this }} (northstar_id, device_id, journey_begin_ts)"
```
rapala61, line 321: @sheyd Is the `IF NOT EXISTS` required here? also, does the space before the column names matter?

sheyd: @rapala61 Also, I don't think we need the `IF NOT EXISTS` since we're using the `table` materialized mode? But I'm happy to add in `IF NOT EXISTS` as a safety measure.

sheyd: Actually, just double-checked, and for tables that don't have `IF NOT EXISTS`, we don't have duplicate indexes, so we can leave this as is.

sheyd:
@rapala61 This is time critical we get this out today. I've pinged Sohaib about getting comments added in here, which we can do as another PR. Can I get a ✔️ so I can test out everything runs in Prod?

sheyd:
@rapala61 Commented SQL now in place. 🗡

sohaibhasan:
@rapala61 @sheyd I accidentally posted a version of the file where the sql was referring to the tables themselves rather than the {{ref}} stuff!

sheyd:
@sohaibhasan D'OH! Do you have a commented `ref` version?

sohaibhasan:
@sheyd Yup, I just swapped them out in the ticket

rapala61:
Good catch! 🦅

PR #991, opened 2020-04-07
remove bertly component bc of lack of visibility into all web traffic

sohaibhasan:
#### What's this PR do?
* Remove bertly dependency on ovrd recipient funnel
* ovrd_recipient_funnel.sql
* run sql
* Related to https://www.pivotaltracker.com/story/show/171886321


blisteringherb:
@sohaibhasan would you mind updating the version in the setup.py file so that the dependency graph can be updated?

PR #996, opened 2020-04-08
OVRD and Voter Reg dbt docs

blisteringherb:
#### What's this PR do?
This adds the docs for the OVRD funnels and for the Voter Reg Quiz funnel. 

@sohaibhasan or @sheyd I need input on the description, and the field definitions. Feel free to add those directly to the PR, or throw them in a comment and I can put them in. Thanks!

https://www.pivotaltracker.com/n/projects/2076969/stories/172012754

### DIFF
--- a/quasar/dbt/models/schema.md
+++ b/quasar/dbt/models/schema.md
@@ -680,7 +680,7 @@ Date when user signed up for newsletter
 {% enddocs %}
 
 {% docs journey_begin_ts %}
-
+Earliest recorded page visit
 {% enddocs %}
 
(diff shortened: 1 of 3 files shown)


sheyd:
@blisteringherb I'm happy to take some guesses here, but I think @sohaibhasan should provide the initial pass, and I'll add on whatever's necessary after that. Also happy if @sohaibhasan wants to tag me on fields he's unclear on and do the footwork to investigate them.

sohaibhasan:
(by the way I added more fields to both of these funnels today 😬

PR #1007, opened 2020-04-20
Removing unique index from MEL field, which shouldn't impact anything.

sheyd:
#### What's this PR do?
* Fixes a unique index conflict issue in `member_event_log`.
https://github.com/DoSomething/quasar/compare/mel-index-update?expand=1#diff-9464235ad31ae62317534b9ac5104902R222
QA.
https://www.pivotaltracker.com/story/show/172412083


sheyd:
@rapala61 I looked at the downstream tables and the raw tables, and didn't see any impact. I was curious to get this into QA to test a bit more, but I couldn't find the underlying raw table collision event. I think it's worth capturing if we have duplicate records as well and why they might be happening. 

I could do some more manual testing by running the raw query in Prod the index as well in another schema to see if I get some more collisions/more info.

rapala61:
Not sure if super helpful, but I found this link about how some index corruption might be hiding the offending duplicate row https://stackoverflow.com/questions/45315510/postgres-could-not-create-unique-index-key-is-duplicated. 

IMO if we do have duplicates in the MEL data, we should address that as a bug rather than updating the assumption that now MEL data is not constrained to be unique. However, not sure how much of a blocker this patch is right now.

Just a heads up. After implementing testing. This is a change that would definitely break a test and would need to be updated.

sheyd:
@rapala61 I'm going to move my next steps into ticket, but basic version is going to manually run SQL and re-create indexes, and see if it breaks in the same way, which would lead us to more troubleshooting steps.

PR #1011, opened 2020-04-22
Updates campaign_info model with new metadata provided by Eri

rapala61:
#### What's this PR do?
It merges the updates to the Campaign Info model provided by @eriqa. It adds the columns:
- `action_types`
- `post_types`
- `online_offline`
- `scholarship`
- `post_types`

Tested manually against QA:
```
rpacas@Rafaels-MBP
 in   ~/Desktop/repos/q/quasar/dbt (campaign-info-metadata-updates) $ dbt run -m campaign_info --profile qa
Running with dbt=0.14.2
Found 37 models, 135 tests, 2 snapshots, 0 analyses, 117 macros, 0 operations, 0 seed files, 3 sources

13:24:42 | Concurrency: 4 threads (target='default')
13:24:42 | 
13:24:42 | 1 of 2 START table model public.campaign_info........................ [RUN]
13:24:42 | 2 of 2 START table model public.campaign_info_international.......... [RUN]
13:24:45 | 1 of 2 OK created table model public.campaign_info................... [SELECT 736 in 1.83s]
13:24:45 | 2 of 2 OK created table model public.campaign_info_international..... [SELECT 177 in 2.22s]
13:24:45 | 
13:24:45 | Finished running 2 table models in 4.66s.

Completed successfully

Done. PASS=2 WARN=0 ERROR=0 SKIP=0 TOTAL=2
```
- https://www.pivotaltracker.com/story/show/172238840

### DIFF
--- a/quasar/dbt/models/campaign_info/campaign_info.sql
+++ b/quasar/dbt/models/campaign_info/campaign_info.sql
@@ -1,57 +1,50 @@
-with
+WITH
 --Campaigns with Online/Offline Post Components
-campaign_online as (
-  select campaign_id, count(distinct online) as online_count, min(case when online=true then 'Online' else 'Offline' end) as min_online
-  from {{ ref('post_actions') }}
-  group by 1
-)
-,
-campaign_online_combo as (
-    select campaign_id, case when online_count>1 then 'Both' else min_online end as online_offline
-    from campaign_online
-)
-,
+campaign_online AS (
+  SELECT campaign_id, count(DISTINCT online) AS online_count, min(CASE WHEN online=true THEN 'Online' ELSE 'Offline' END) AS min_online
+  FROM {{ ref('post_actions') }}
+  GROUP BY 1
+),
+campaign_online_combo AS (
+    SELECT campaign_id, CASE WHEN online_count>1 THEN 'Both' ELSE min_online END AS online_offline
+    FROM campaign_online
+),
 --Campaigns and Action Types
-campaign_action as (
-  select campaign_id, action_type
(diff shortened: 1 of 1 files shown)


quasar/dbt/models/campaign_info/campaign_info.sql
```diff
@@ -1,3 +1,58 @@
...
+campaign_action as (
+  select campaign_id, action_type
+  from {{ ref('post_actions') }}
+  where action_type is not null and action_type<>''
+  group by 1, 2
+)
+,
+--Campaigns and Action Types Combined
+campaign_action_combo as (
+    select campaign_id, string_agg(action_type, ' / ' order by action_type) as action_types
```
rapala61, line 24: @eriqa is there a reason why `action_types` and `post_types` use the `/` separator instead of a `,`? We use `,`s in `campaign_action_type` and `campaign_cause_type`.

eriqa: @rapala61 Just from experience when combining 2 values into 1 filed while keeping visual clue of the 2 separate values, I've used the / (forward slash) separator instead of a , (comma) to avoid potential troubles when the data gets exported into CVS, and the commas in the values could be misinterpreted as field separators. But if it's best practice to use the comma, I'm OK with changing it. Thanks.

rapala61: @eriqa that's a really valid point. I've seen that happen a lot with Text type columns! For consistency, I think we should stick to one or the other. I'll update to use `,`s for consistency with the rest of the columns in the table.

PR #1013, opened 2020-04-24
refactor ovrd recipient funnel

sohaibhasan:
#### What's this PR do?
* Implements the ability to show page views as part of the funnel
* Complete refactor to make the funnel to easier to extend to new funnel checkpoints in the future
* quasar/dbt/models/user_journey/ovrd_recipient_funnel.sql
* Run in QA
* We will soon bring OVRD pages into our ecosystem and off instapage, this allows us to implement snowplow tracking. This change sets us up so these events will flow into the funnel as soon as they start appearing
* The refactoring also sets us up to more easily extend this funnel if the OVRD recipient flow changes in the future
* There is a corresponding Looker PR
https://www.pivotaltracker.com/story/show/171886321

### DIFF
--- a/quasar/dbt/dbt_project.yml
+++ b/quasar/dbt/dbt_project.yml
@@ -250,8 +250,8 @@ models:
           alias: ovrd_recipient_funnel
           materialized: table
           post-hook:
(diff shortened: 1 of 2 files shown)


blisteringherb:
I made some updates to the indexes since they were still referencing `northstar_id`

sohaibhasan:
Oh right! Good catch

PR #1018, opened 2020-05-01
Adding User Engagement tables to DBT

blisteringherb:
#### What's this PR do?
This PR adds the User Engagement DBT code. I'm taking this opportunity to use an intermediate schema for the staging tables. 
#### What are the relevant tickets?
https://www.pivotaltracker.com/n/projects/2076969/stories/172239031

PR #1025, opened 2020-05-05
Fix issues with OVRD and MEL from merge

blisteringherb:
#### What's this PR do?
This PR fixes the broken MEL index and incorrect fields on the OVRD recipient funnel indexes. These got broken from merging in this PR. https://github.com/DoSomething/quasar/pull/1018/files
https://www.pivotaltracker.com/n/projects/2076969/stories/172679594

### DIFF
--- a/quasar/dbt/dbt_project.yml
+++ b/quasar/dbt/dbt_project.yml
@@ -270,7 +270,7 @@ models:
           alias: member_event_log
           materialized: table
           post-hook:
-           - "CREATE UNIQUE INDEX ON {{ this }}(timestamp, northstar_id, event_id)"
+           - "CREATE INDEX ON {{ this }}(timestamp, northstar_id, event_id)"
            - "CREATE INDEX ON {{ this }}(northstar_id, timestamp)"
            - "GRANT SELECT ON {{ this }} TO dsanalyst"
            - "GRANT SELECT ON {{ this }} TO looker"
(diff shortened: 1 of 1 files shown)


quasar/dbt/dbt_project.yml
```diff
@@ -301,8 +301,8 @@ models:
           alias: ovrd_recipient_funnel
           materialized: table
           post-hook:
-            - "CREATE UNIQUE INDEX IF NOT EXISTS ovrd_r_index ON {{ this }}(northstar_id)"
-            - "CREATE INDEX ON {{ this }}(northstar_id)"
+            - "CREATE UNIQUE INDEX IF NOT EXISTS ovrd_r_index ON {{ this }}(user_id)"
```
sheyd, line 304: I don't think we need to do `IF NOT EXISTS` in DBT index creation since we use `table` materialization mode?

blisteringherb: Have we changed the outlook on this? Looks like we decided to add this clause back in August? https://github.com/DoSomething/quasar/pull/807#pullrequestreview-280891759

sheyd: Hmmm, good point. But we have `CREATE INDEX` on other tables and we don't have runaway duplicate indexes. I'm guessing I was wrong and this was in relation to using Indexes when using `incremental` materialization mode. :thinking: It won't do any harm as is, so feel free to merge for expediency, but I'd like to figure out what the answer on this one is!

sheyd: Approved so we can get this out.

blisteringherb: On the upside, when and if we go to convert to incremental builds, we'll already have this code in place.

PR #1032, opened 2020-05-15
Cio email bounced email event

blisteringherb:
#### What's this PR do?
This PR adds the `email_event` and `email_bounced` models to DBT.
https://www.pivotaltracker.com/n/projects/2328687/stories/172866953


rapala61:
@blisteringherb I think you have to merge my branch into yours and set the PR to merge into mine. instead of master.

blisteringherb:
> @blisteringherb I think you have to merge my branch into yours and set the PR to merge into mine instead of master.

@rapala61 I changed the base to your branch, I think that might do the trick?

quasar/dbt/models/cio/email_event.sql
```diff
@@ -0,0 +1,20 @@
...
+	event #>>'{data, link_id}' as link_id,
+	event ->> 'event_id' AS event_id,
+	TO_TIMESTAMP(CAST(event ->> 'timestamp' AS INTEGER)) AS "timestamp",
+	event #>>'{data, variables, campaign, id}' as cio_campaign_id,
+	event #>>'{data, variables, campaign, name}' as cio_campaign_name,
+	event #>>'{data, variables, campaign, type}' as cio_campaign_type,
+	event #>>'{data, event_type}' as event_type
+FROM
+    {{ source('cio', 'event_log') }} cel
+WHERE event->event_type = 'email_bounced'
```
rapala61, line 20: I think you want this be something like 

```
NOT IN ('email_sent', 'email_bounced', 'customer_subscribed', 'customer_unsubscribed')
```

blisteringherb: Sorry! I missed that one. I updated that and used the list from the queue processing.
```
        email_event = {
            'email_bounced',
            'email_converted',
            'email_opened',
            'email_unsubscribed'
        }
```
https://github.com/DoSomething/quasar/blob/2d3d1ec0b0ce2b060255deb23c9104ed66d280f7/quasar/cio_queue.py#L188

quasar/dbt/models/cio/email_event.sql
```diff
@@ -0,0 +1,20 @@
...
+	event #>>'{data, link_id}' as link_id,
+	event ->> 'event_id' AS event_id,
+	TO_TIMESTAMP(CAST(event ->> 'timestamp' AS INTEGER)) AS "timestamp",
+	event #>>'{data, variables, campaign, id}' as cio_campaign_id,
+	event #>>'{data, variables, campaign, name}' as cio_campaign_name,
+	event #>>'{data, variables, campaign, type}' as cio_campaign_type,
+	event #>>'{data, event_type}' as event_type
+FROM
+    {{ source('cio', 'event_log') }} cel
+WHERE event->event_type IN ('email_bounced', 'email_converted', 'email_opened', 'email_unsubscribed')
```
rapala61, line 20:
```suggestion
WHERE event->event_type NOT IN ('email_bounced', 'email_converted', 'email_opened', 'email_unsubscribed')
```

quasar/dbt/dbt_project.yml
```diff
@@ -388,3 +388,17 @@ models:
             - "CREATE INDEX IF NOT EXISTS {{ this.schema ~ '_' ~ this.table }}_email_id_customer_id_timestamp ON {{ this }}(email_id, customer_id, timestamp)"
             - "GRANT SELECT ON {{ this }} to dsanalyst"
             - "GRANT SELECT ON {{ this }} to looker"
+        email_bounced:
+          alias: cio_email_bounced
+          materialized: table
+          post-hook:
+            - "CREATE INDEX IF NOT EXISTS {{ this.schema ~ '_' ~ this.table }}_event_id ON {{ this }}(email_id, customer_id, "timestamp")"
```
rapala61, line 395: I think we also want to update the name of the indexes to match the columns.

quasar/dbt/models/cio/email_event.sql
rapala61: Oh! Got it... This is partially my bad. email bounced should not be in this list though.

blisteringherb: It's listed as one of the events we currently included in that `email_event` table. Are we switching up that logic?

rapala61: @sheyd is this a 🐛 or is intentional? We have an `email_bounced` table, do we want this event also included in the `email_event` table?

sheyd: Intentional from when Sohaib set it up this way. I'd say for simplicity, let's leave the existing table structures as they are, and talk about better layout once we're back to ingesting normally.

quasar/dbt/dbt_project.yml
```diff
@@ -388,3 +388,17 @@ models:
             - "CREATE INDEX IF NOT EXISTS {{ this.schema ~ '_' ~ this.table }}_email_id_customer_id_timestamp ON {{ this }}(email_id, customer_id, timestamp)"
             - "GRANT SELECT ON {{ this }} to dsanalyst"
             - "GRANT SELECT ON {{ this }} to looker"
+        email_bounced_new:
+          alias: cio_email_bounced_new
```
rapala61, line 392: One approach I've taken is not to add `_new` to the alias. I think it will help in the long run not to surface this implementation detail in the reference names. Let me know if you think this is a better way so I can update my tables!

quasar/dbt/models/cio/email_bounced_new.sql
```diff
@@ -0,0 +1,19 @@
...
+    event ->> 'event_id' AS event_id,
+    TO_TIMESTAMP(CAST(event ->> 'timestamp' AS INTEGER)) AS "timestamp",
+    event #>>'{data, variables, campaign, id}' as cio_campaign_id,
+    event #>>'{data, variables, campaign, name}' as cio_campaign_name
+FROM
+    {{ source('cio', 'event_log') }} cel
+WHERE event #>>'{data, event_type}' = 'email_bounced'
+UNION ALL
+SELECT *
+FROM {{ source('cio', 'email_bounced') }}
```
rapala61, line 19: Should this source reference `email_bounced_old`? Same for `email_event_old`.

rapala61:
@blisteringherb I also found a few hardcoded strings referencing the legacy cio tables in a few other tables other than MEL, e.g. user activity and users table models.

blisteringherb:
@rapala61 I've updated those files where the references were hardcoded.

quasar/dbt/models/cio/email_bounced_new.sql
blisteringherb: @rapala61 Now that I'm thinking about it a bit more, should we be renaming those since they're in a different schema? That also makes me wonder if it's useful to be using the `_new` suffix for the derived tables since they're separated in a new schema and they're built a different way.

quasar/dbt/dbt_project.yml
rapala61: @blisteringherb could we change the aliases to `cio_email_bounced_event` and `cio_email_event` for consistency with the other 2 tables?

quasar/dbt/models/cio/email_bounced_new.sql
rapala61: @blisteringherb that makes a lot of sense. I removed the `_new` suffix for the other two tables based on this comment. Thanks!

rapala61: However, I left the sources pointing to the `_old` suffixed tables. I think it helps clear any confusion that they should not be used because are deprecated.

quasar/dbt/models/users_table/cio_latest_status.sql
```diff
@@ -5,12 +5,12 @@ SELECT
...
 			ELSE 'customer_subscribed' END) AS event_type,
 	max(cio."timestamp") AS "timestamp"
-FROM cio.customer_event cio
+FROM {{ ref('cio_email_event_new') }} cio
 INNER JOIN 
 	(SELECT 
 		ctemp.customer_id,
 		max(ctemp."timestamp") AS max_update
-	FROM cio.customer_event ctemp
+	FROM {{ ref('cio_email_event_new') }} ctemp
```
rapala61, line 13: @blisteringherb Just a heads up that removing the `_new` suffix in the aliases will require updates on references like this one.

quasar/dbt/models/cio/email_bounced_new.sql
blisteringherb: A little while ago we created a `historical_analytics` schema. Do you think we should move the old `cio` tables there and then use something like:
```
sources:
  - name: cio
    schema: '{{ env_var("CIO_WEBHOOK_EVENTS_SCHEMA") }}'
    tables:
      - name: event_log
  - name: historical_analytics
    schema: '{{ env_var("HISTORICAL_ANALYTICS") }}'
    tables:
      - name: cio_email_bounced
    tables:
      - name: cio_email_event
```
And then reference them with:
`{{ source('historical_analytics', 'cio_email_event') }}`

quasar/dbt/models/user_activity/user_activity.sql
```diff
@@ -60,7 +60,7 @@ email_unsub AS (
 	    LAST_VALUE(event_type) OVER (
 		PARTITION BY customer_id ORDER BY "timestamp", event_type
 		RANGE BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING) AS last_status
-	    FROM cio.customer_event
+	    FROM {{ ref('customer_event') }}
```
rapala61, line 63: @blisteringherb I'm making this change in my PR, you don't need to include it in your or it will override!

quasar/dbt/models/users_table/cio_latest_status.sql
```diff
@@ -5,12 +5,12 @@ SELECT
 			THEN 'customer_unsubscribed' 
 			ELSE 'customer_subscribed' END) AS event_type,
 	max(cio."timestamp") AS "timestamp"
-FROM cio.customer_event cio
+FROM {{ ref('cio_email_event') }} cio
```
rapala61, line 8: @blisteringherb same for the changes in this file. They are being applied in https://github.com/DoSomething/quasar/pull/1031. This would override them.

quasar/dbt/models/cio/email_bounced_new.sql
rapala61: Just read this. Great idea. This is definitely something we should do. I'll create a cleanup card to address this. Thanks @blisteringherb.

rapala61: Pivotal Card: https://www.pivotaltracker.com/story/show/172923044

PR #1047, opened 2020-06-01
Snowplow incremental

blisteringherb:
#### What's this PR do?
This PR adds the logic to incrementally build the Snowplow models. This sets a DBT variable that uses an ENV var with an interval that allows us to adjust the increment window for Snowplow updates. I preferred to use something like max(event_datetime), but Postgres doesn't allow aggregate functions in `WHERE` clauses. With our time savings on running incremental builds, we should be able to rely on `DISTINCT` to remove duplicates.

I've set the env var in QA in production:
`export SNOWPLOW_INTERVAL="'24 HOURS'"`

This has been tested with the following commands:
```
└─▪ dbt run -m phoenix_events --profile prod --target prod
Running with dbt=0.16.1
Found 48 models, 135 tests, 2 snapshots, 0 analyses, 127 macros, 0 operations, 0 seed files, 9 sources

19:41:48 | Concurrency: 1 threads (target='prod')
19:41:48 |
19:41:48 | 1 of 7 START incremental model public.snowplow_base_event............ [RUN]
19:54:24 | 1 of 7 OK created incremental model public.snowplow_base_event....... [INSERT 0 270057 in 756.00s]
19:54:24 | 2 of 7 START incremental model public.snowplow_payload_event......... [RUN]
19:56:22 | 2 of 7 OK created incremental model public.snowplow_payload_event.... [INSERT 0 282891 in 117.65s]
19:56:22 | 3 of 7 START incremental model public.snowplow_raw_events............ [RUN]
20:00:01 | 3 of 7 OK created incremental model public.snowplow_raw_events....... [INSERT 0 270194 in 219.08s]
20:00:01 | 4 of 7 START incremental model public.snowplow_phoenix_events........ [RUN]
20:03:16 | 4 of 7 OK created incremental model public.snowplow_phoenix_events... [INSERT 0 270057 in 195.10s]
20:03:16 | 5 of 7 START incremental model public.snowplow_sessions.............. [RUN]
06:00:02 | 5 of 7 OK created incremental model public.snowplow_sessions......... [INSERT 0 42168 in 35805.82s]
06:00:02 | 6 of 7 START incremental model public.phoenix_events_combined........ [RUN]
06:10:10 | 6 of 7 ERROR creating incremental model public.phoenix_events_combined [ERROR in 607.92s]
06:10:10 | 7 of 7 START incremental model public.phoenix_sessions_combined...... [RUN]
06:13:49 | 7 of 7 OK created incremental model public.phoenix_sessions_combined. [INSERT 0 14081099 in 219.81s]
06:13:50 |
06:13:50 | Finished running 7 incremental models in 37923.76s.

Completed with 1 error and 0 warnings:

Database Error in model phoenix_events_combined (models/phoenix_events/phoenix_events_combined.sql)
  duplicate key value violates unique constraint "pec_unique"
  DETAIL:  Key (event_datetime, event_name, event_id)=(1970-01-01 02:48:02.705+00, visit, 5cefe3bc14dd3800043c84d3) already exists.
  compiled SQL at ../../docs/run/ds_dbt/phoenix_events/phoenix_events_combined.sql

Done. PASS=6 WARN=0 ERROR=1 SKIP=0 TOTAL=7
```
```
└─▪ dbt run -m phoenix_events.phoenix_events_combined --full-refresh --profile prod --target prod
Running with dbt=0.16.1
Found 48 models, 135 tests, 2 snapshots, 0 analyses, 127 macros, 0 operations, 0 seed files, 9 sources

20:37:34 | Concurrency: 1 threads (target='prod')
20:37:34 |
20:37:34 | 1 of 1 START incremental model public.phoenix_events_combined........ [RUN]
21:42:15 | 1 of 1 OK created incremental model public.phoenix_events_combined... [SELECT 98302511 in 3880.96s]
21:42:15 |
21:42:15 | Finished running 1 incremental model in 3887.93s.

Completed successfully

Done. PASS=1 WARN=0 ERROR=0 SKIP=0 TOTAL=1
```
```
└─▪ dbt run -m phoenix_events.phoenix_sessions_combined --full-refresh --profile prod --target prod
Running with dbt=0.16.1
Found 48 models, 135 tests, 2 snapshots, 0 analyses, 127 macros, 0 operations, 0 seed files, 9 sources

21:50:44 | Concurrency: 1 threads (target='prod')
21:50:44 |
21:50:44 | 1 of 1 START incremental model public.phoenix_sessions_combined...... [RUN]
21:51:46 | 1 of 1 OK created incremental model public.phoenix_sessions_combined. [SELECT 23210403 in 62.49s]
21:51:46 |
21:51:46 | Finished running 1 incremental model in 69.19s.

Completed successfully

Done. PASS=1 WARN=0 ERROR=0 SKIP=0 TOTAL=1
```
https://www.pivotaltracker.com/story/show/173082160
#### Questions:
```

### DIFF
--- a/quasar/dbt/dbt_project.yml
+++ b/quasar/dbt/dbt_project.yml
@@ -34,8 +34,6 @@ models:
   ds_dbt:
       # Applies to all files under models/phoenix_events/
       phoenix_events:
-        vars:
-            run_interval: "to_timestamp({{ run_started_at.timestamp()|int }}) - INTERVAL {{env_var('SNOWPLOW_INTERVAL')}}"
         snowplow_base_event:
           alias: snowplow_base_event
           materialized: incremental
@@ -58,7 +56,7 @@ models:
           alias: snowplow_phoenix_events
           materialized: incremental
           post-hook:
-            - "CREATE INDEX IF NOT EXISTS {{ this.schema ~ '_' ~ this.table }}_spe_{{ run_started_at.timestamp()|int }} ON {{ this }}(event_datetime, event_name, event_id)"
+            - "CREATE INDEX IF NOT EXISTS {{ this.schema ~ '_' ~ this.table }}_spe ON {{ this }}(event_datetime, event_name, event_id)"
             - "CREATE INDEX IF NOT EXISTS spe_session_id ON {{ this }} (session_id)"
             - "GRANT SELECT ON {{ this }} TO looker"
             - "GRANT SELECT ON {{ this }} TO dsanalyst"
@@ -72,7 +70,7 @@ models:
           alias: phoenix_events_combined
           materialized: incremental
           post-hook:
-            - "CREATE INDEX IF NOT EXISTS {{ this.schema ~ '_' ~ this.table }}_pec_{{ run_started_at.timestamp()|int }} ON {{ this }}(event_datetime, event_name, event_id)"
+            - "CREATE INDEX IF NOT EXISTS {{ this.schema ~ '_' ~ this.table }}_pec ON {{ this }}(event_datetime, event_name, event_id)"
             - "CREATE INDEX IF NOT EXISTS pec_session_id ON {{ this }} (session_id)"
             - "GRANT SELECT ON {{ this }} TO looker"
             - "GRANT SELECT ON {{ this }} TO dsanalyst"
--- a/quasar/dbt/dbt_project.yml
+++ b/quasar/dbt/dbt_project.yml
@@ -72,7 +72,7 @@ models:
           alias: phoenix_events_combined
           materialized: incremental
           post-hook:
-            - "CREATE UNIQUE INDEX IF NOT EXISTS pec_unique ON {{ this }}(event_datetime, event_name, event_id)"
+            - "CREATE INDEX IF NOT EXISTS {{ this.schema ~ '_' ~ this.table }}_pec_{{ run_started_at.timestamp()|int }} ON {{ this }}(event_datetime, event_name, event_id)"
             - "CREATE INDEX IF NOT EXISTS pec_session_id ON {{ this }} (session_id)"
             - "GRANT SELECT ON {{ this }} TO looker"
             - "GRANT SELECT ON {{ this }} TO dsanalyst"
--- a/quasar/dbt/dbt_project.yml
+++ b/quasar/dbt/dbt_project.yml
@@ -34,41 +34,43 @@ models:
   ds_dbt:
       # Applies to all files under models/phoenix_events/
       phoenix_events:
+        vars:
+            run_interval: "to_timestamp({{ run_started_at.timestamp()|int }}) - INTERVAL {{env_var('SNOWPLOW_INTERVAL')}}"
         snowplow_base_event:
           alias: snowplow_base_event
-          materialized: table
+          materialized: incremental
           post-hook:
            - "CREATE INDEX IF NOT EXISTS event_index ON {{ this }}(event_id)"
            - "GRANT SELECT ON {{ this }} TO dsanalyst"
         snowplow_payload_event:
           alias: snowplow_payload_event
-          materialized: table
+          materialized: incremental
           post-hook:
            - "CREATE INDEX IF NOT EXISTS payload_event_id ON {{ this }}(event_id)"
            - "GRANT SELECT ON {{ this }} TO dsanalyst"
         snowplow_raw_events:
           alias: snowplow_raw_events
-          materialized: table
+          materialized: incremental
           post-hook:
             - "CREATE INDEX IF NOT EXISTS raw_event_id ON {{ this }}(event_id)"
             - "GRANT SELECT ON {{ this }} TO dsanalyst"
         snowplow_phoenix_events:
           alias: snowplow_phoenix_events
-          materialized: table
+          materialized: incremental
           post-hook:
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: quasar/dbt/dbt_project.yml [hunk 1]
```diff
@@ -34,49 +34,51 @@ models:
   ds_dbt:
       # Applies to all files under models/phoenix_events/
       phoenix_events:
+        vars:
+            run_interval: "to_timestamp({{ run_started_at.timestamp()|int }}) - INTERVAL {{env_var('SNOWPLOW_INTERVAL')}}"
```
rapala61, line 38: Minor, but would love to get a comment above this line that points to this: https://jinja.palletsprojects.com/en/2.11.x/templates/#filters or even this https://stackoverflow.com/questions/19161093/convert-integer-to-string-jinja/19162679#19162679

quasar/dbt/dbt_project.yml
```diff
@@ -34,49 +34,51 @@ models:
...
           post-hook:
             - "CREATE INDEX IF NOT EXISTS raw_event_id ON {{ this }}(event_id)"
             - "GRANT SELECT ON {{ this }} TO dsanalyst"
         snowplow_phoenix_events:
           alias: snowplow_phoenix_events
-          materialized: table
+          materialized: incremental
           post-hook:
-            - "CREATE INDEX IF NOT EXISTS {{ this.schema ~ '_' ~ this.table }}_spe_unique_{{ run_started_at.timestamp()|int }} ON {{ this }}(event_datetime, event_name, event_id)"
+            - "CREATE INDEX IF NOT EXISTS {{ this.schema ~ '_' ~ this.table }}_spe_{{ run_started_at.timestamp()|int }} ON {{ this }}(event_datetime, event_name, event_id)"
```
rapala61, line 61: Since the name changes dynamically each time `{{ this.schema ~ '_' ~ this.table }}_spe_{{ run_started_at.timestamp()|int }}` , wouldn't this index be created on every run?

quasar/dbt/dbt_project.yml
```diff
@@ -34,49 +34,51 @@ models:
...
             - "CREATE INDEX IF NOT EXISTS sps_landing ON {{ this }}(landing_datetime, landing_page)"
             - "GRANT SELECT ON {{ this }} TO looker"
             - "GRANT SELECT ON {{ this }} TO dsanalyst"
         phoenix_events_combined:
           alias: phoenix_events_combined
-          materialized: table
+          materialized: incremental
           post-hook:
-            - "CREATE UNIQUE INDEX IF NOT EXISTS pec_unique ON {{ this }}(event_datetime, event_name, event_id)"
+            - "CREATE INDEX IF NOT EXISTS {{ this.schema ~ '_' ~ this.table }}_pec_{{ run_started_at.timestamp()|int }} ON {{ this }}(event_datetime, event_name, event_id)"
```
rapala61, line 75: Same question as https://github.com/DoSomething/quasar/pull/1047/files#r433262817

sheyd:
@blisteringherb for
>I preferred to use something like max(event_datetime), but Postgres doesn't allow aggregate functions in WHERE clauses

Can you go into a little more detail? I'm coming at this fresh, so sorry if this is a dumb/repetitive question, but guessing that you used something [similar to this](https://docs.getdbt.com/docs/building-a-dbt-project/building-models/configuring-incremental-models/#filtering-rows-on-an-incremental-run) to attempt using `max(timestamp)`. If that's the case, the `max` clause would be part of the `SELECT` sub_query which should work. I ran a test query in QA to validate this assumption (and it ran with no issues):

```
SELECT *
FROM ft_snowplow."event" e 
WHERE collector_tstamp > (SELECT max(dvce_tstamp) FROM ft_snowplow."event" e2);
```

Would the clause looking like this work potentially?:
```
{% if is_incremental() %}
-- this filter will only be applied on an incremental run
WHERE s.event_datetime >= (SELECT max(event_datetime) FROM {{ this.table }})
{% endif %}
```

Given that you've tested and the above probably doesn't work, my only note would be to set interval to `25` hours instead of `24` since we might miss a small number of records assuming small diffs in job runtime actual kickoffs (e.g. one night's job starts as 00:00:00 and next night 00:00:05, etc).

sheyd:
> * Instead of using an interval in a variable. What do you think of adding a column like `_dbt_incremental_synced` and calculating a `max(_dbt_incremental_synced)` on that to inform the timestamp of new data? I'm just concerned that we run into a situation where we need to stop the Jenkins job for more than 24 hours. We just need to keep track of this limitation for this initial version. NOTE: Maybe this is what Sena also refers to [here](https://github.com/DoSomething/quasar/pull/1047#issuecomment-636894997)
Yep, that's exactly my concern as well.

> * What do you think if, for the incremental SQL clause, we use Macros instead? https://docs.getdbt.com/docs/writing-code-in-dbt/macros/

I like the portability and cleanness of `macros` for this 🔪 .

blisteringherb:
I made updates to the date calculation for these incremental builds and used the `max()` function. It seems that when you run `max()` in a `WHERE` clause, you have to specify the table along with the field name in that `SELECT`subquery or else Postgres complains that you can't run aggregate functions in a `WHERE`. I ran a full refresh last night, and going forward we should be able to run this on our daily schedule and only get the new records in. 

@rapala61 I know you had some concerns about the previous approach. Do you feel more comfortable with this one?

PR #1048, opened 2020-06-01
Adds Cio events import script

rapala61:
#### What's this PR do?
- It adds script to import Cio events from a file.
- https://www.pivotaltracker.com/story/show/173068221

### DIFF
--- a/quasar/misc/import-cio-events-script.py
+++ b/quasar/misc/import-cio-events-script.py
@@ -22,11 +22,12 @@ import json
 import math
 import os
 import sys
+import signal
 import time
 from colorama import init, Fore, Back
 from datetime import datetime
 from halo import Halo
-from sqlalchemy import (create_engine, event, MetaData,
+from sqlalchemy import (create_engine, DDL, event, MetaData,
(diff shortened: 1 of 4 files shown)


rapala61:
@sheyd @blisteringherb any thoughts or questions about the scope of the script (as comments inside the script) are welcome.

sheyd:
I like the non-destructive nature here, and seems like a good approach! Maybe make the `1000` record call a variable to be able to tweak on-demand in case API changes happen?

quasar/misc/import-cio-events-script.py
```diff
@@ -0,0 +1,201 @@
...
+    return datetime.today().strftime('%Y%m%d')
+
+
+def get_username_hash():
+    return adler32(bytes(getpass.getuser().encode('utf-8')))
+
+
+def get_import_table_name():
+    # We are hashing and using the script executor's username as part
+    # of the table name to prevent a collition with a parallel import
```
sheyd, line 85: `collision` instead of `collition`.

PR #1064, opened 2020-07-27
Use 'ft_bertly' to ingest Bertly 2.0 clicks.

DFurnes:
#### What's this PR do?
This pull request updates our Bertly DBT model to use `ft_bertly`, which is where Fivetran ingests clicks for Bertly 2.0, rather than `bertly`, our old Amazon DMS schema. Both have the same structure, so it should be a simple change.

I swapped the source schema in `bertly_clicks.sql`. Is there anything else I need to change?

I also updated the lockfile (based on changes in #1059) in response to this notice when I ran `pipenv install`:

```
$ pipenv install

Pipfile.lock (21c72f) out of date, updating to (1b168b)…
Locking [dev-packages] dependencies…
Building requirements...
Resolving dependencies...
✔ Success!
```

We [discussed](https://www.pivotaltracker.com/story/show/173797616/comments/216334170) moving the `bertly.clicks` table to the `historical_analysis` schema before running this, so we could be sure that it was correctly ingesting the new `ft_bertly` schema instead.

This will allow us to see clicks generated by Bertly 2.0 (which writes click logs to a S3 bucket, rather than a PostgreSQL instance) in Looker, so we can continue to see how those impact our monthly active user metrics.

[#173797616](https://www.pivotaltracker.com/story/show/173797616)

#### Screenshots (if appropriate)
N/A

#### Questions:
- Is there any other way I should test or validate this locally? I set up a local environment following the README, but I wasn't sure if there was any way I should go about manually testing that DBT runs properly using that...
- Does this automatically deploy to QA? How do I run DBT there? (Jenkins job?)


rapala61:
We were just discussing this over the sync. There are a few similar ways to tackle this, I think the easiest is to update the ENV var for the [Bertly source](https://github.com/DoSomething/quasar/blob/master/quasar/dbt/models/sources.yml#L5-L8) so it points to the new schema. This way there is no code changes. There is a case for creating a new source instead, in that case, add it to the `sources.yml` file and update the same model you updated here with the correct source name.

DFurnes:
Oh! I wish I'd know this was on the sync agenda.

That looks like the right way to handle this. How do we manage environment variables for this application?

DFurnes:
Also, do we need the `Pipfile.lock` changes from this PR, or shall I discard?

rapala61:
@DFurnes sorry about the sync agenda! I don't remember how we ended up talking about the topic, but it wasn't part of the agenda. I added the "Jenkins-QA/PROD SSH config" note in LastPass to help you SSH into the EC2 instances to change the ENV var. I might just need you to send me your public SSH over slack. We can sync over Slack about it if you want.

blisteringherb:
> Also, do we need the `Pipfile.lock` changes from this PR, or shall I discard?

@DFurnes I think we could go either way. I don't think leaving those changes in will hurt things, but OTOH maybe we should keep the PR "clean" if it doesn't require these updates?

DFurnes:
Ok! I only asked because it seems like we might've forgotten to commit them before.

I'll just close this pull request out.

PR #1072, opened 2020-08-06
Renaming, adding timeout, and env var for limit

blisteringherb:
#### What's this PR do?
* Adds `timeout_s` param to the client setup
  - The default was 1s which was causing the script to timeout often. The timeouts are intermittent, so it may take some time to make sure this works.
* Modifies the API limit to use and ENV var
  - So we can make these updates without a code change
* Renames the file for consistency
* Adds an error condition for ReadTimeout

I've tested it manually locally.
https://www.pivotaltracker.com/story/show/173799216

### DIFF
--- a/quasar/contentful_metadata.py
+++ b/quasar/contentful_metadata.py
@@ -3,6 +3,7 @@ import contentful
 from pprint import pprint
 from .sa_database import Database
 import jsonpickle
+from urllib3.exceptions import ReadTimeoutError
 
 db = Database()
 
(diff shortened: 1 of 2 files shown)


quasar/contentful_metadata.py
```diff
@@ -42,6 +43,8 @@ def _save_campaign(campaign):
         pprint('{contentful_id} ran into a type error: {error}'.format(contentful_id=campaign.id, error=te))
     except (AttributeError) as e:
         pprint('{contentful_id} missing an attribute: {error}'.format(contentful_id=campaign.id, error=e))
+    except (ReadTimeout) as to:
+        pprint('{contentful_id} ran into a type error: {error}'.format(contentful_id=campaign.id, error=to))
```
rapala61, line 47: I think it would be helpful to mention this is a timeout error here. e.g. "ran into a timeout error"

rapala61:
Also so it's more clear, what is the `timeout_s` now after the update?

blisteringherb:
@rapala61 I've set that to 30 seconds on prod.

PR #1096, opened 2020-10-02
Schema tests for snowplow phoenix events model 

rapala61:
#### What's this PR do?
I took a look at the `snowplow_phoenix_events` model to implement schema tests. I found that we should not test this model. The only difference between this model and its parent `snowplow_raw_events` is the existence of the `campaign_name` column. We join `campaign_info` to pull this value. We could do just that in the parent node and save us the space.

```
relid   |schema_name|table_name             |table_size|indexes_size|total_size|
--------|-----------|-----------------------|----------|------------|----------|
23059596|public     |snowplow_phoenix_events|21 GB     |8997 MB     |29 GB     |
```
NOTE: Eliminating this model will require that we update old references and point them to `snowplow_raw_events`.

The changes in this PR, aside from comments, were just indentation and sorting related.

- https://www.pivotaltracker.com/story/show/173910879

### DIFF
--- a/quasar/dbt/models/phoenix_events/schema.yml
+++ b/quasar/dbt/models/phoenix_events/schema.yml
@@ -126,111 +126,111 @@ models:
   - name: snowplow_raw_events
     description: Table combining snowplow_base_event and snowplow_payload_event into rich raw data source for all clickstream info
     columns:
-        - name: block_id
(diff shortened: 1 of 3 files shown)


ghangas-ds (approved):
Agree with you that we should remove the unnecessary step and join campaign in the previous model and remove the redundancy!

PR #1098, opened 2020-10-06
Updating voter reg logic

blisteringherb:
#### What's this PR do?
This PR updates the voter reg logic for reportbacks.

https://www.pivotaltracker.com/story/show/175119147


ghangas-ds (approved):
@blisteringherb i changed my mind on this. let's go ahead and merge this now. There are internal DS campaigns being named with the word 'vote' and it will become a bigger problem if those keep getting classified as voter reg reportbacks.

PR #1100, opened 2020-10-15
Decrease data model build times

rapala61:
#### What's this PR do?
- It updates DBT to v0.18.1. It makes our `dbt_project.yml` conf file compatible with the new features in `config-version: 2`.
- DRYs up select access to DBT models by creating the `grant_select_to` macro and using it in the `post-hook` property for each model.
- Refactors `device_campaign` to take less time to build
- https://www.pivotaltracker.com/story/show/175166734

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -32,7 +32,7 @@ pipenv = "*"
 sh = "*"
 contentful = "*"
 python-dateutil = "*"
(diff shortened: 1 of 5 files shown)


eriqa (commented):
Decreasing model rebuilt times is ok with me

blisteringherb (approved):

PR #1103, opened 2020-11-02
Adding snapshot export to s3

blisteringherb:
#### What's this PR do?
This PR adds the code to start the export of the snapshots and push those into the `dosomething-quasar-archive` s3 bucket.

https://www.pivotaltracker.com/story/show/174829369

### DIFF
--- a/quasar/aws_s3.py
+++ b/quasar/aws_s3.py
@@ -5,9 +5,9 @@ import collections
 s3 = boto3.client('s3')
 rds = boto3.client('rds')
 
-# Backup the newest snapshot each day, check to make sure we're not backing it up twice
-# Make sure the backup was successful
-# Put them in Glacier in 30 days
+# Backup the newest snapshot each day, check to make sure we're not backing it
+# up twice.
+
 
 def list_backups():
     s3_objects = s3.list_objects_v2(Bucket='dosomething-quasar-archive', Delimiter="/")
@@ -15,6 +15,7 @@ def list_backups():
     for folder in s3_objects['CommonPrefixes']:
         folders.append(folder['Prefix'][:-1])
     return folders
+
 
 def list_snapshots():
     response = rds.describe_db_snapshots(DBInstanceIdentifier='quasar-prod')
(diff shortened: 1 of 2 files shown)


rapala61:
@blisteringherb I'll wait for the linter fixes before reviewing.

quasar/aws_s3.py
```diff
@@ -0,0 +1,75 @@
...
+    rds_snapshots = list_snapshots()
+    s3_backups = list_backups()
+    # Check which of these snapshots is not saved in the s3 bucket
+    # Export RDS snapshots to S3.
+    to_backup = list(set(rds_snapshots) - set(s3_backups))
+
+    task_status = check_backup_status()
+
+    backup_slots = 5 - task_status['running']  # StartExportTask only allows for 5 concurrent backups
+    if backup_slots > 5:
```
rapala61, line 61: @blisteringherb quick question. Maybe i'm misunderstanding. If there are only 5 slots for concurrent backup tasks. `backup_slots` should be between 0 and 5. Shouldn't we check here that `backup_slots < 5` instead?

blisteringherb: You're totally right! Had a bit of a brain fart there.

quasar/aws_s3.py
```diff
@@ -0,0 +1,73 @@
...
+            i = 0
+            if i <= backup_slots:
+                for snapshot in to_backup:
+                    rds.start_export_task(ExportTaskIdentifier=snapshot,  # S3 folder name
+                                          SourceArn=rds_snapshots[snapshot],  # RDS Snapshot ARN
+                                          S3BucketName=os.environ.get('EXPORT_S3_BUCKET_NAME'),
+                                          IamRoleArn=os.environ.get('EXPORT_ROLE_ARN'),
+                                          KmsKeyId=os.environ.get('EXPORT_KMS_ID'))
+                    i += 1
+        except Exception as e:
```
rapala61, line 72: @blisteringherb I see that we are catching the exception here. Is there a way we could handle this so that the script exits with a failure status when this occurs?

blisteringherb: @rapala61 Could you clarify what you mean by exit with a failure status?

rapala61: Sorry for the delay! @blisteringherb If this script fails in this try...catch block. Would the Jenkins job fail? If I understand correctly, since we are "catching" the exception and printing it, it would not fail, and would mark the job build as successful. I think we want it to fail so we can be notified and is marked as failure in the jobs history.

Context: We had this similar behavior in the Cio consumer job back in the day. It contributed in hiding the "dropped" events due to some schema inconsistencies. The job would just "print" the error and continue, not alerting us of the failure.

blisteringherb: Good point, I added a `sys.exit(-1)` to fail the job if there's an exception.

rapala61 (approved):
Thank you for the minor changes. LGTM! 🚀

PR #1112, opened 2021-01-13
Feature/expose hours spent

ghangas-ds:
#### What's this PR do?

- Adds the hours_spent field for volunteer credits to the posts and reportbacks model
- Also makes a minor update to the name of the field is_time_commitment --> time_commitment (since it's not a boolean flag)

Tested locally in QA:

```
Params-MBP:dbt pghangas$ pipenv run dbt run -m campaign_activity.posts --profile qa
Running with dbt=0.18.1
Found 49 models, 88 tests, 3 snapshots, 0 analyses, 143 macros, 0 operations, 0 seed files, 24 sources

15:48:17 | Concurrency: 4 threads (target='default')
15:48:17 | 
15:48:17 | 1 of 1 START table model public.posts................................ [RUN]
15:48:47 | 1 of 1 OK created table model public.posts........................... [SELECT 2265977 in 30.52s]
15:48:48 | 
15:48:48 | Finished running 1 table model in 31.62s.

Completed successfully
```

- https://www.pivotaltracker.com/story/show/176384079

### DIFF
--- a/quasar/dbt/models/campaign_activity/posts.sql
+++ b/quasar/dbt/models/campaign_activity/posts.sql
@@ -24,7 +24,7 @@ SELECT
 	a.quiz AS is_quiz,
 	a.reportback AS is_reportback,
 	a.scholarship_entry AS is_scholarship_entry,
-	a.time_commitment AS is_time_commitment,
+	a.time_commitment AS time_commitment,
 	a.volunteer_credit AS is_volunteer_credit,
 	pd.location,
 	pd.northstar_id AS northstar_id,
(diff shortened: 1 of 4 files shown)


rapala61 (commented):
@ghangas-ds Looks great! just have 2 minor questions.

Since you are also making an update to an existing column name (`is_time_commitment`). Did you check if the column has any dependencies in downstream models? are there any documentation updates associated with this column change?

ghangas-ds:
> @ghangas-ds Looks great! just have 2 minor questions.
> 
> Since you are also making an update to an existing column name (`is_time_commitment`). Did you check if the column has any dependencies in downstream models? are there any documentation updates associated with this column change?

Ah, good question. I did check and I didn't see any downstream dependencies (we don't currently bring into reportbacks) and I think the docs don't need to be updated... since this is just changing the alias name ??

rapala61:
> > @ghangas-ds Looks great! just have 2 minor questions.
> > Since you are also making an update to an existing column name (`is_time_commitment`). Did you check if the column has any dependencies in downstream models? are there any documentation updates associated with this column change?
> 
> Ah, good question. I did check and I didn't see any downstream dependencies (we don't currently bring into reportbacks) and I think the docs don't need to be updated... since this is just changing the alias name ??

Now that I check, that column is not even [documented](https://github.com/DoSomething/quasar/blob/23bfe1f044ca7da7182dc6450152ea5816e11535/quasar/dbt/models/campaign_activity/schema.yml#L23) for the model 😅 . If you want, document it in this same PR, but I will approve as is.

rapala61 (approved):
🥇 !

PR #1116, opened 2021-01-25
Adds data test to monitor invalid user snapshots

rapala61:
#### What's this PR do?
- Adds data test to monitor invalid user snapshots

- https://www.pivotaltracker.com/story/show/176196749


ghangas-ds (approved):
Logic looks good. And very easy to read and interpret this code!

PR #1120, opened 2021-02-23
Removes Zendesk DBT model

rapala61:
#### What's this PR do?
It removes v1 of the Zendesk model. We need to do a quick review of the configurations before submitting V2.


ghangas-ds (approved):
Makes sense!

PR #1131, opened 2021-08-17
Adding env example file.

weerd:
This PR adds a simple `.env.example` file with the full list of ENV variables that the project may utilize. 

There may be some variables that are now deprecated and not actually used in the project, but we'll need to remove those as we have a better idea of where/if they are used!

For some of the variables related to table fields, I did not include the values since I wanted to confirm it would be "ok" to have them in source control before I add them. I don't think it's a concern to have them, but wanted to double-check.


weerd:
There is already a `quasar-sample.env` file which doesn't match the standard naming and hasn't been updated in 3 years. So I think it'd be safe to remove with now adding in this new `.env.example` but again, wanted to double-check! 😅

DFurnes (approved):
Fantastic!

DFurnes:
@weerd Want to add a gitignore for `.env` while you're in here?

DFurnes:
> There is already a `quasar-sample.env` file which doesn't match the standard naming and hasn't been updated in 3 years. So I think it'd be safe to remove with now adding in this new `.env.example` but again, wanted to double-check! 😅

Hmm yeah, I agree! We can always reference via git history if we need to look something up.

weerd:
Ok! After discussing, we went ahead and added the table related field/values in the specified env variables since that would be a pain to look up and does not pose a security issue.

